PR #1, opened 2018-03-21
.gitignore

Banbeucmas:
We are Open Source, let's fucking go

P.S: Copy pasta from MCMMO .gitignore wooo

PR #2, opened 2018-03-21
Do some improvements to the codebase

DoNotSpamPls:
Don't merge this yet, it still has some work to be done, here's what I've done so far:

- Shade bStats instead of having the class in the plugin
- Use POM placeholders in plugin.yml and config.yml (get it, because... it's a placeholder plugin :smirk:
okay i'm out)
- Use Bukkit's config load system instead of making the config on the fly.


Banbeucmas:
https://github.com/PlaceholderAPI/PlaceholderAPI/pull/1

extendedclip:
Thank you for the contribution, means a lot :sparkling_heart:

Issue #5, opened 2018-03-27
Change ecloud commands to only require permission node rather than Op


extendedclip:
As stated here: https://www.spigotmc.org/threads/placeholderapi.61918/page-91#post-2944172
Some may find it annoying that ecloud commands require op to be executed. I personally feel it is safer but I guess perm checking is just as secure.  Any feedback is appreciated.

RustyDagger:
I find having to login to my console view to run the command time consuming I can access every thing else from in game with out being Op, If my account was breached they could do far harder things to undo than install a few modules. 

some may ask why I am not Op Op tends to just get handed every thing this can include boots to xp gain on some plugins that's undesirable

DoNotSpamPls:
I think making it a permission would make it more versatile, allowing to reject access even for opped people. It can just be added to plugin.yml with `default: op` as redundancy

Banbeucmas:
Pretty sure this might create some inconvience for some LuckPerm users who `allows-Op` is set to `false` if this doesn't get implemented.

extendedclip:
Thanks @DoNotSpamPls for the PR.

T

PR #6, opened 2018-03-27
Check for permissions instead of isOp()

DoNotSpamPls:
Related: #5

### DIFF
--- a/src/main/java/me/clip/placeholderapi/commands/PlaceholderAPICommands.java
+++ b/src/main/java/me/clip/placeholderapi/commands/PlaceholderAPICommands.java
@@ -74,7 +74,7 @@ public class PlaceholderAPICommands implements CommandExecutor {
 				"&b/papi reload",
 				"&fReload the config settings");
 
-				if (s.isOp()) {
+				if (s.hasPermission("placeholderapi.ecloud")) {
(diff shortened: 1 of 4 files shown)


src/main/java/me/clip/placeholderapi/commands/PlaceholderAPICommands.java
```diff
@@ -110,7 +110,7 @@ public boolean onCommand(CommandSender s, Command c, String label, String[] args
 				return eCloud.onCommand(s, c, label, args);
 				
 			} else if (args[0].equalsIgnoreCase("enablecloud")) {
-				if (!s.isOp()) {
+				if (!s.hasPermission("placeholderapi.ecloud")) {
```
Banbeucmas: Should the permission be specific? `placeholderapi.ecloud.toggle` or something.

extendedclip:
So we doing this or what

src/main/java/me/clip/placeholderapi/commands/PlaceholderAPICommands.java
DoNotSpamPls: I thought of doing that, but to be honest, if you have enough power to install extensions on the server, you most likely have full rights (or close to full rights)

PR #21, opened 2018-05-25
Time Edit

PrayaAmadigaPitasa:


src/main/java/me/clip/placeholderapi/configuration/PlaceholderAPIConfig.java
```diff
@@ -63,4 +66,16 @@ public String booleanFalse() {
 	public String dateFormat() {
 		return plugin.getConfig().getString("date_format");
 	}	
+	
+	public String getTime(TimeFormat timeFormat) {
+		final FileConfiguration config = plugin.getConfig();
+		
+		switch (timeFormat) {
```
Sxtanna: This isn't a very extensible format to write code in.

Observing this code block, you will notice that the config path is directly reflected in the name of the associated enum constant. This entire block of code could be effectively one line.

src/main/java/me/clip/placeholderapi/util/TimeUtil.java
```diff
@@ -20,143 +20,60 @@
...
-		}	
-	}
+public class TimeUtil {
 	
-	public static String getTime(int seconds) {
+	public static final String getRemaining(int seconds, TimeFormat type) {
+		final PlaceholderAPIPlugin plugin = (PlaceholderAPIPlugin) JavaPlugin.getProvidingPlugin(PlaceholderAPIPlugin.class);
+		final PlaceholderAPIConfig config = plugin.getPlaceholderAPIConfig();
+		final String time = config.getTime(type);
+		final int days = seconds / 86400; seconds = seconds - (days * 86400);
```
Sxtanna: I would highly suggest taking advantage of the TimeUnits made specifically to do these sort of conversions. It would make the code a lot less verbose, and increase general readability in a function like that which tends to becomes a "wall" of code.

src/main/java/me/clip/placeholderapi/util/TimeUtil.java
```diff
@@ -20,143 +20,60 @@
...
+	public static final String getTime(int seconds) {
+		final PlaceholderAPIPlugin plugin = (PlaceholderAPIPlugin) JavaPlugin.getProvidingPlugin(PlaceholderAPIPlugin.class);
+		final PlaceholderAPIConfig config = plugin.getPlaceholderAPIConfig();
+		final String time = config.getTime(type);
+		final String timeDays = getTime(TimeFormat.DAYS);
+		final String timeHours = getTime(TimeFormat.HOURS);
+		final String timeMinutes = getTime(TimeFormat.MINUTES);
+		final String timeSeconds = getTime(TimeFormat.SECONDS);
+		final StringBuilder timeBuilder = new StringBuilder();
+		final int days = seconds / 86400; seconds = seconds - (days * 86400);
```
Sxtanna: `seconds -=` instead of `seconds = seconds -`.

Also, putting these statements on the same line as an initialization of a variable greatly decreases readability.

extendedclip:
This commit isn’t needed right now so I’m going to close it. We will be changing this until in papi 3.

PR #26, opened 2018-06-18
Add loadbefore for tab plugin

615283:
This means that the plugin no longer has to be manually reloaded after server starts when using PAPI placeholders within the configuration.

### DIFF
--- a/src/main/resources/plugin.yml
+++ b/src/main/resources/plugin.yml
@@ -2,6 +2,7 @@ name: ${project.name}
 main: me.clip.placeholderapi.PlaceholderAPIPlugin
 version: ${project.version}
 author: [extended_clip]
+loadbefore: [TAB]
 description: ${project.description}
 permissions:
     placeholderapi.*:


extendedclip:
Rather than loading up the PAPI plugin.yml with tons of plugins it should load before, don't you think it may be wise to have the depending plugin softdepend or depend on PlaceholderAPI which would give the same result? I think that would be better. That way PlaceholderAPI is not worried about load order as the depending plugin can ensure it loads after PAPI.

615283:
I was gonna do that but TAB isn't open-source afaik

615283:
Compromise: in config, we have a section that you can put plugins to load before PAPI then have something in the onEnable dealing with this. Thoughts @extendedclip?

mikroskeem:
Bug "TAB" plugin author with this, or just patch the plugin.yml in "TAB" plugin =)

615283:
@mikroskeem see my second above message. (`I was gonna do that but TAB isn't open-source afaik`)

mikroskeem:
You can literally unzip the jar, modify plugin.yml and repack it

mikroskeem:
If the original author is not open enough, then making changes locally is the best you can do.

PR #55, opened 2018-11-23
1.13 tab-complete issue fix

LogGits:
This *should* fix 1.13.2 tab-complete list from showing commands that a player doesn't have permission to use. We could also default it to true and allow people to negate the permission to stop it from showing in tab-complete for people who shouldn't have access to it.


darbyjack:
This PR is being closed due to the v3.0 rewrite which won't contain this anymore.

PR #61, opened 2018-12-27
Load at startup

Phoenix616:
By default plugins are loaded later. (post world) This complicates hooking into PlaceholderAPI from other plugins that load directly at startup as PlaceholderAPI isn't available yet at that point.

Besides that I feel like such a library should be available as early as possible as it itself doesn't depend on anything else but lots of other plugins and plugins depending on them might depend on it.

### DIFF
--- a/src/main/resources/plugin.yml
+++ b/src/main/resources/plugin.yml
@@ -4,6 +4,7 @@ version: ${project.version}
 author: [extended_clip]
 description: ${project.description}
 api-version: 1.13
+load: STARTUP
 permissions:
     placeholderapi.*:
         description: ability to use all commands


extendedclip:
Thanks @Phoenix616

mibby:
@extendedclip This seems to break a lot of things for me. Using ChatControl-Pro with PAPI expansions and MobArenaExpansion plugin.

link

Downgrading to dev 58, all is back to normal.

Phoenix616:
@mibby this might be because the expansions are registered before the plugin is actually loaded, but as far as I can tell this case [is checked](https://github.com/PlaceholderAPI/PlaceholderAPI/blob/master/src/main/java/me/clip/placeholderapi/expansion/ExpansionManager.java#L114) and then cached for registering [when the plugin is actually loaded](https://github.com/PlaceholderAPI/PlaceholderAPI/blob/master/src/main/java/me/clip/placeholderapi/PlaceholderListener.java#L80-L89)?

mibby:
Should I create a ticket for this issue? Because this commit is quite breaking in my environment. :(

extendedclip:
Na just use the current release until I work this issue out. I have an idea on what I can do to fix the issue which won't break any compatibility with plugins being loaded before or after papi but I need to try it out and see.

PR #97, opened 2019-04-21
No more daddy

NDVelsen:
Bye daddy.

### DIFF
--- a/src/main/resources/plugin.yml
+++ b/src/main/resources/plugin.yml
@@ -40,4 +40,4 @@ permissions:
 commands:
    placeholderapi:
      description: PlaceholderAPI command
-     aliases: [clipsplaceholderapi, papi, daddy]
+     aliases: [clipsplaceholderapi, papi]


Andre601:
That is honestly a useless PR.
If you really don't want `daddy` as a subcommand (which is sort of a inside/translation joke), then get the non-daddy version from here: https://github.com/Sauilitired/PlaceholderAPI

NDVelsen:
I quote

> If anyone submits a PR removing the alias,. I guess I will remove it. It was just an inside joke we enjoy on discord so I thought it would be cool to actually add. @mrfloris sorry if this upset you...

Andre601:
From where should that quote be?

NDVelsen:
https://github.com/PlaceholderAPI/PlaceholderAPI/commit/589712f7e4290803e407e7b6253478363895756a

Andre601:
But whatever the reason is, this is not really needed.
If you don't like the command, don't use it. Or if you're a dev, make and use your own version. But a PR for just that is basically useless.
When it would have more changes in general then sure, but just for removing this command that didn't bother anyone so far (At least I never saw anyone complaining about its existance).

Also clip said "I guess" which doesn't mean he would clearly remove it.
Sorry, but no. If you make a PR with more useful changes, then go ahead, but this one here doesn't improve anything.

NDVelsen:
You can simply remove "daddy" out of the plugin.yml after downloading the jar, no need to fork.
I like how you say that my PR isn't really needed, but this alias isn't really needed as well. It's just inappropriate and uncalled for.

Andre601:
image

NDVelsen:
That's literally what I've said. 
**I like how you say that my PR isn't really needed, but this alias isn't really needed as well. It's just inappropriate and uncalled for.

Care to explain why daddy is needed?**

Andre601:
Again: This is not a issue here.
So far I didn't see any user that came and directly complained about the alias being there.
In what way is it inappropriate? It is a translation of the word `papi` which means `daddy` in Spanish (And also in German btw).
People know that `papi` is short for `PlaceholderAPI`, but why not being able to make a simple running gag?
People never complained about PermissionsEx turning into PermissionSex on 1st of april.
This PR was and still is useless since it only changes one small thing that isn't even that annoying (Again: I so far didn't see any person complaining about this alias)
I will lock this PR now because it doesn't help to complain about this because it is "inappropriate".

Funny-cube:
Please Keep conversations relevant development of the plugin.

//Locked and Pruned

PR #107, opened 2019-05-03
3.0 changes

Andre601:
[3.0-branch]: https://github.com/PlaceholderAPI/PlaceholderAPI/tree/3.0

### About
This Draft PR acts as a central place for discussions about the planned v3.0 of PlaceholderAPI.

### What will change?
PlaceholderAPI will be a complete rewrite, meaning it will have **breaking changes** to plugins and expansions alike.

There will be a 2.x release where we allow devs to update their plugins and expansions, to use the new API while still supporting the current one for the time being.

### Contribution
If you have ideas, suggestions or even own Code-changes, feel free to comment here or make a new PR. Make sure to have the latest version from the [3.0-branch] and that no other PR has your changes when trying to submit your own PR.

### DIFF
--- a/src/main/java/me/clip/placeholderapi/configuration/PlaceholderAPIConfig.java
+++ b/src/main/java/me/clip/placeholderapi/configuration/PlaceholderAPIConfig.java
@@ -54,9 +54,9 @@ public class PlaceholderAPIConfig {
 		c.addDefault("check_updates", true);
 		c.addDefault("cloud_enabled", true);	
 		c.addDefault("cloud_allow_unverified_expansions", false);
-		c.addDefault("boolean.true", "yes");
-		c.addDefault("boolean.false", "no");
-		c.addDefault("date_format", "MM/dd/yy HH:mm:ss");
+		c.addDefault("defaults.boolean_true", "true");
+		c.addDefault("defaults.boolean_false", "false");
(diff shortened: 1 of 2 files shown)


extendedclip:
So are we moving the entire plugin to the new package name com.placeholderapi in 3.0?

darbyjack:
I think that would follow the best formatting and practices since we own that domain.

Andre601:
> So are we moving the entire plugin to the new package name com.placeholderapi in 3.0?

I think we could do the new stuff there and keep the old path as a thing for people to update their plugins to the new API while it still works with the old one.
We can then later completely switch to the new one.

Andre601:
Updated the first comment.

Andre601:
Addiitional plan that was just recently discussed (and probably still will be) is the change of registering the expansions (jars) to a similar system like spigot/bukkit has with its `plugin.yml`

The idea I had was having a `papi.yml` that contains all important information that you normally register/set through the different get methods like `getIdentifier()`, `getAuthor()`, etc.

This is a (rough) sketch of how the file could look like:
```yaml
Author: Someone
version: 1.0.0
identifier: something

# Would be used to choose if PlaceholderAPI should f.e. delay the registration.
depends: Vault

main: at.helpch.example.Example

# Idea for per-version methods
versions:
  '1.8.8': at.helpch.example.1_8_8.Example
  '1.12.2': at.helpch.example.1_12_2.Example
```

PR #131, opened 2019-06-10
Formatting, spacing, final modifiers, and a few code fixes

Draycia:
- Use append instead of string concatenation in StringBuilder
- Remove unnecessary casts in the PlaceholderAPI class
- Add final modifiers to some fields
- Ensure line width doesn't exceed a certain amount
- Add spacing mostly where appropriate


src/main/java/me/clip/placeholderapi/PlaceholderAPI.java
```diff
@@ -374,18 +398,18 @@ public static boolean unregisterPlaceholderHook(Plugin plugin) {
   }
 
   public static String setPlaceholders(Player p, String text) {
-    return setPlaceholders((OfflinePlayer) p, text, PLACEHOLDER_PATTERN);
+    return setPlaceholders(p, text, PLACEHOLDER_PATTERN);
```
Andre601: You sure this won't cause issues?

Draycia: Player extends OfflinePlayer so there shouldn't be any problems, the cast is redundant.

I'll triple check everything works when I get time to today.

Draycia: Everything seems to be working fine, OtherPlayer and Player placeholders function as they should.

darbyjack:
Thanks for this. I'll look over it with clip in the coming days and hopefully, this will be pulled into the next update!

extendedclip:
Will follow up with this pr when I get home from vacation.

PR #141, opened 2019-07-07
Add api-version to plugin.yml

Leomixer17:

## Please read

### Type

- [ ] Other: __________


darbyjack:
Adding in that will prevent lower versions from working.

Andre601:
> Adding in that will prevent lower versions from working.

Maybe have it on 1.13?
1.12 and older don't care about this since they don't have the check for it and 1.14 certainly wouldn't complain about this since it's more or less the same API like 1.13 but with additional Material names and such...
You should really add AT LEAST `api-version: 1.13` since it's an annoyance to constantly have this warning in console on startup.

PR #195, opened 2019-10-09
Port PAPI to the NukkitX platform

GMatrixGames:

## Please read

### Type

- [x] Other: Internal and external 

This pull request implements a port of PlaceholderAPI to the Bedrock Edition server platform, NukkitX.

Though following a similar API, there were some compromises and some code that ended up not being used or unable to be used with the other platform.

With this PR, to allow for a single branch instead of one for each version, I've converted the existing maven setup, to a multi-module one. (It still compiles successfully). 

The only caveat of this, is the ecloud system, though implemented, cannot be used as the current v2 API endpoint would end up causing issues with expansions for different platforms, having the same name. Because of this, in the config.yml file, I have pre-disabled ecloud downloads, and pre-enabled unverified expansions.

**Reviews welcome, it's my first PR to PlaceholderAPI** _Closes #166_


Andre601:
How exactly does the ecloud not allow expansions for nukkit?

GMatrixGames:
I should probably re-phrase that. I was speaking of the fact that there are some spigot plugins that are on the Nukkit platform as well, so in theory, it would have to have 2 expansions on the ecloud with the same name

Andre601:
I think having a separate endpoint or *category* on the ecloud to determine if the expansion is for Spigot or Nukkit servers would imo be an idea.
Then PAPI could, when it is on sponge, access the Nukkit endpoint of the ecloud.

GMatrixGames:
That's the type of thing I was going for, I'm probably going to edit the PR
message to do a better explanation. (It was really late at night for me)

aBooDyy:
Would it be possible to use one jar (expansion) for both platforms? Instead making 2 for each plugin and changing how the eCloud work.

Andre601:
> Wouldn't it be possible to use one jar (expansion) for both platforms? Instead making 2 for each plugin and changing how the eCloud work.

I think changing the ecloud would be kinda recommended as f.e. downloading the jar for the wrong platform could cause severe issues.
And I think that as long as Nukkit uses a different kind of system (like a different kind of plugin registry) that won't interfere with Spigots would this sure be possible by having two separate main classes handling the stuff.

GMatrixGames:
@aBooDyy Would you suggest I open a PR to some of the expansions that are made by the PlaceholderAPI team, to add the Nukkit stuff so that if this gets pulled, we have pre-done placeholders?

aBooDyy:
I would say PR when this gets pulled. If you want to start working on them now, it's up to you.

PiggyPiglet:
Great job if this works, I haven’t tested or looked at the code yet. However, ever since your pr, we’ve been discussing the topic in our staff chats and are envisioning a fully platform independent papi; this is a step in the right direction.

Regarding the expansions, there most certainly won’t be multiple implementations for different servers for the one expansion. I think we’re going to go the route of providing wrappers for all the essential portions of each server implementation through papi itself, thus a fully independent api for expansions to use. This won’t be 100% fool proof, and however slight it may be, there’s still the possibility that an expansion may need a niche feature papi doesn’t yet have coverage for, in which case the expansion will become platform specific.

Just a disclaimer, this is all just speculation and brain storming for papi 3, which is currently in the works. There’s no eta, and it’s not guaranteed any of the above will actually make it in.

Until something like the above is implemented though, I see no problem with having your version available, even via this org - it’s not my call though; you’ll have to wait for clip.

GMatrixGames:
I'm looking forward to the final decision! I saw a few requests in other discords, and the feature request on this repo, and figured that it would be a good idea to port it over for the Bedrock Community (as the current PlaceholderAPI they have is pretty bad).

LemmoTresto:
I've had a glance at the code and I've instantly seen that you are not using abstraction and definitely not following the DRY principle.

You have literally copy-pasted the existing codebase and edited it to make it work with Nukkit, this is definitely not the way to go.

Instead, you should make abstractions and use a 'common'/'core' codebase to extend from.
This common codebase will be very easy to use and packs as much code as possible without making it platform dependent. Then all you need to do to support various platforms is making the required listeners/commands for said platform and interact with the common codebase.

I highly suggest revising this as I see no way this will get merged in its current state. 

It's a step in the right direction and this has broadened our look on PlaceholderAPI and the demand for it on other platforms.

GMatrixGames:
I was originally planning on using the "common module" setup, but wasn't sure if that was the way the project managers would want it setup. I will work on refractoring out classes that can be shared between projects.

GMatrixGames:
@LemmoTresto I have a feeling I should close this PR, and make a new one, as with the new workspace, it changes a lot more than just porting the plugin

GamerDaniel3000:
@GMatrixGames  👍

GMatrixGames:
Moved to #204

PR #295, opened 2020-04-26
Save Cacheable expansions data on shutdown

iGabyTM:
### Type

- [ ] Other: __________ 

Save Cacheable expansions data on shutdown


darbyjack:
According to the PR merged yesterday, this change was included in it.

PR #323, opened 2020-06-11
Update bStats

Luracasmus:
### Type


darbyjack:
Doesn't really need to be updated. Not game-breaking.

PR #333, opened 2020-06-25
1.16

darbyjack:
Compiled and tested under 1.16


RedyAu:
Does a70b9f8 fix PAPI for 1.16?

darbyjack:
This was merged into main already.

PR #340, opened 2020-07-01
Performance Improvements

CryptoMorin:
### Type
- [ ] Other: __________ 

### Description
Performance improvements for most of the classes.
The major improvement is the PlaceholderReplacer class which isn't replaced with the main RegEx replacer. I'm not sure if you want to use this, because it's a little hard to understand. If you want I can remove PlaceholderReplacer but it's much better than compiled RegEx. This method is tested, but there is a chance that it might break with very rare conditions.
I understand if you don't want to use this new method. Let me know how you like it.

Other changes are just pure improvements with no risks. Some are really minor improvements, but why not.
And yes, I did some benchmarking and it indeed improves performance, not significantly of course.


darbyjack:
Hi there, 

we just merged a major PR into our live branch. If you could implement your changes on top of that, we can look into merging this! 

Thanks!

CryptoMorin:
What are you going to do about the PlaceholderReplacer? Do you want to keep this new method or do you want to stick to the RegEx replacer?

PiggyPiglet:
> What are you going to do about the PlaceholderReplacer? Do you want to keep this new method or do you want to stick to the RegEx replacer?

Have you benchmarked the difference between your solution, and the compiled regex? I honestly wouldn't be surprised if the regex was faster, compiled regex is very fast in java.

extendedclip:
Yes, before the recent changes we made, we did test your method and I think we should move forward with using your method over regex. 

We did have a good talk on discord about your method and I do remember someone saying there was an issue with the "closure" detection but either way lets move forward with it for internal parsing of placeholders. just have to keep the regex patterns and other methods ex: containsPlaceholders etc.

CryptoMorin:
> Have you benchmarked the difference between your solution, and the compiled regex? I honestly wouldn't be surprised if the regex was faster, compiled regex is very fast in java.

Of course. It's x5 times faster. It's just technically impossible for RegEx (compiled or not) to beat manual character loop.
B-99

Then I'll replace the internal replacer with this and keep the existing methods. I'll do some more testing to see if there are any bugs in special cases.

PR #346, opened 2020-07-07
Add cmdparse command

stonar96:

## Please read

### Type

- [x] Other: Added a command, which an end-user **can** use. 

This PR adds the command `/papi cmdparse <player|me> <command with placeholders>`, which parses an expansion (command) and dispatches the result as command.

### DIFF
--- a/src/main/java/me/clip/placeholderapi/commands/CommandHandler.java
+++ b/src/main/java/me/clip/placeholderapi/commands/CommandHandler.java
@@ -24,6 +24,7 @@ public final class CommandHandler implements CommandExecutor {
             new UnregisterCommand(),
             new ReloadCommand(),
             new BcParseCommand(),
+            new CmdParseCommand(),
             new ParseCommand(),
             new ParseRelCommand(),
 
--- a/src/main/java/me/clip/placeholderapi/commands/command/BcParseCommand.java
+++ b/src/main/java/me/clip/placeholderapi/commands/command/BcParseCommand.java
@@ -35,7 +35,7 @@ public final class BcParseCommand extends Command {
             }
         }
 
-        String parse = StringUtils.join(args, " ", 2, args.length);
+        String parse = StringUtils.join(args, " ", 1, args.length);
         Msg.broadcast("&r" + PlaceholderAPI.setPlaceholders(player, parse));
     }
 }
--- a/src/main/java/me/clip/placeholderapi/commands/command/CmdParseCommand.java
+++ b/src/main/java/me/clip/placeholderapi/commands/command/CmdParseCommand.java
@@ -1,0 +1,41 @@
+package me.clip.placeholderapi.commands.command;
+
+import me.clip.placeholderapi.PlaceholderAPI;
+import me.clip.placeholderapi.commands.Command;
+import me.clip.placeholderapi.util.Msg;
+import org.apache.commons.lang.StringUtils;
+import org.bukkit.Bukkit;
+import org.bukkit.OfflinePlayer;
+import org.bukkit.command.CommandSender;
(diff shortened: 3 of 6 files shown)


Andre601:
I don't see an actual benefit in this. This could easly be made into a separate expansion.

stonar96:
And where is the benefit of the existing commands `/papi parse` and `/papi bcparse`? I don't see an actual reason not to add this. Furthermore my additions are much more abstract and general than the existing parse and broadcast commands. For example you can even replace `/papi bcparse me ...` by `/papi cmdparse me tellraw @a ...` and do much more things I can't even think of.

Andre601:
> And where is the benefit of the existing commands `/papi parse` and `/papi bcparse`? I don't see an actual reason not to add this. Furthermore my additions are much more abstract and general than the existing parse and broadcast commands. For example you can even replace `/papi bcparse me ...` by `/papi cmdparse me tellraw @a ...` and do much more things I can't even think of.

The parse and bcparse commands are used to find out, what the placeholders would return based on the provided player. Parse only does this for you and bcparse does it for every player on the server and broadcasts it.
I really can't see a reason why a command should execute another command.
I didn't see anyone use the bcparse command either, so my guess is, that this command isn't really used. The parse command is useful, as it helps you debug stuff and find out, how a specific text would look like, when applied to a player.

stonar96:
I accept your opinions. However I still think this is very general and useful thing.

> I really can't see a reason why a command should execute another command.

This is actually a very common thing and there are multiple examples. Even in Minecraft there is the `/execute ... run ...` command. Other examples are `sudo ...` or `screen -dmS name ...`.

> I didn't see anyone use the bcparse command either, so my guess is, that this command isn't really used. The parse command is useful, as it helps you debug stuff and find out, how a specific text would look like, when applied to a player.

True and `cmdparse` would be a command which can be actually useful apart of debugging. Many plugins have configurabe commands to execute on a specific event. By merging this PR all those plugins implicitely support the PlaceholderAPI. This is pretty useful.

> Additionally would this command add a possible loop-hole, where someone could just do something like `/papi cmdparse me papi cmdparse me papi cmdparse me ...` to loop this pretty much forever.

Not an issue.

That's all I want to say for now. If the dev's decide this shouldn't be added I'm okay with that. But as already said, I think this is useful.

extendedclip:
I don’t see how this would hurt anything being added tbh. My only concern is that we are currently working on merging another pr which modified the command framework so this will most likely have to be rewritten after that is done.

stonar96:
Fine, thanks for the response. Of course I'll wait and I can update this PR once the other one is merged.

stonar96:
I have updated this PR now and fixed the bcparse and parserel command (see changes).

stonar96:
Updated taking into account the latest changes and fixed conflicts.

Sxtanna:
@stonar96 I am currently writing the *hopefully* final version of the papi command system, I can go ahead and integrate this directly into that. Or feel free to create another PR with the changes once I am done.

Sxtanna:
The new parse command is here: https://github.com/PlaceholderAPI/PlaceholderAPI/blob/commands_rewrite/src/main/java/me/clip/placeholderapi/commands/rewrite/impl/CommandParse.java

stonar96:
@Sxtanna thanks I have noticed the new branch already, that's why i didn't fix the current conflicts yet. I guess it would be easier if you go ahead and integrate this directly into that, in case you are okay with that new command. Thanks anyway for the response.

Sxtanna:
@stonar96 all done 😄 0d38618e89a5baa26f7246d40e8b92677013bcd1
image

stonar96:
@Sxtanna Thanks!

PR #357, opened 2020-07-23
Commands rewrite

Sxtanna:
Please accept my humble offering. ❤️


Sxtanna:
It's also probably worth noting that while this is primarily a rewrite of the command system, there are also a couple of changes to the internals of the expansion manager and cloud expansion manager.

These changes were to make the plugin more modern and stable.

knightzmc:
my professional opinion is that this is good

Sxtanna:
> my professional opinion is that this is good

I completely agree with this assessment.

darbyjack:
I really love the way this works how it does because of the way it is.

Andre601:
Needs more /daddy xD

PR #362, opened 2020-07-24
Expansions List updates

Sxtanna:
this.
image


Funny-cube:
The expansion list should be ordered alphabetical to allow users to actually search through the list to find something rather than defaulting to the last update where the list would seem fairly random to the end user.

One solution is to allow the user to specify how it is sorted via some sort of flag/argument

Sxtanna:
Yeah, that's good and all, but that's also not what this PR addresses. This simply reestablishes the rich json messages. Ideation on ordering should be brought up in an issue, and reviewed/PRed from there.

Issue #367, opened 2020-07-24
Return previous papi ecloud list output


SlimeDog:
[Request change (Wiki)]: https://github.com/PlaceholderAPI/PlaceholderAPI/issues/new?template=change_request_wiki.md
[Bug report]: https://github.com/PlaceholderAPI/PlaceholderAPI/issues/new?template=bug_report.md
[issues]: https://github.com/PlaceholderAPI/PlaceholderAPI/issues
[wiki]: https://github.com/PlaceholderAPI/PlaceholderAPI/wiki
[Pull Requests]: https://github.com/PlaceholderAPI/PlaceholderAPI/pulls

## Please read
This template is only for requesting new functions to be added to PlaceholderAPI or for old one to be changed!  
Please use the [Request change (Wiki)] template to ask for additions/changes to the wiki.  
For reporting bugs of PlaceholderAPI use the [Bug report] template.

Also make sure that the [wiki] doesn't already mention such a function/change and that there isn't a issue about this change on the [issues] page.

We recommend to [open a Pull Request][Pull Requests] for making changes to PlaceholderAPI's code.

### Type
> What kind of request is this? (Multiple selections possible)

- [ ] New function for PlaceholderAPI.  
A new function that developers could use.
- [ ] Change to code (Internal).  
Changes to code that won't affect the end-user.
- [x] Change to code (External).  
Changes to code that will affect the end-user (breaks stuff).
- [ ] Other: __________ 

### Info
> What is the change?  
> Please provide as much information (including links, images and code-snippeds) as possible.

Spigot 1.16.1
PlaceholderAPI 2.10.7-DEV-18

Until quite recently (say, 2.10.7-DEV-110), `papi cloud list installed` produced a single line per expansion:
```
>papi ecloud list installed
[09:07:41] [Thread-101/INFO]: [CoreProtect] Processing world data...
[09:07:41] [Server thread/INFO]: Showing expansions for: all installed -- amount: 14 page: 1/2
[09:07:41] [Server thread/INFO]: Green = Expansions you have
[09:07:41] [Server thread/INFO]: Gold = Expansions which need updated
[09:07:41] [Server thread/INFO]: 0: CoordinateTool -- http://api.extendedclip.com/media/PAPI-CoordinateTool-1.1.0.jar
[09:07:41] [Server thread/INFO]: 1: Essentials -- http://api.extendedclip.com/media/PAPI-Expansion-Essentials_CIvJUl0.jar
[...]

PR #369, opened 2020-07-24
Fix missing line break in expansion list command.

Andre601:

## Please read

### Type

- [ ] Other: __________ 

Fixes the issue described [here](https://github.com/PlaceholderAPI/PlaceholderAPI/issues/367#issuecomment-663632503) where the list command doesn't put each expansion on a new list.
I would appreciate if this can be checked to see if this is alright.


SlimeDog:
Confirmed fixed in b22 (and most likely in b21, but I didn't test it).

PR #385, opened 2020-07-28
Add daddy as a command alias

extendedclip:

## Please read

### Type

- [ ] Other: __________ 

Do not delete this PR. Lets talk about it. I want to know why its so bad... I personally like the alias and would use it over /papi any day.

### DIFF
--- a/src/main/resources/plugin.yml
+++ b/src/main/resources/plugin.yml
@@ -10,7 +10,7 @@ description: "@description@"
 commands:
   placeholderapi:
     description: "PlaceholderAPI Command"
-    aliases: ["papi"]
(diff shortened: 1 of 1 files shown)


PiggyPiglet:
So long it's behind a permission like @Sxtanna said, there's no reason this can't be added back.

LemmoTresto:
I'm all for having the alias, afaik it was only removed due to someone complaining. So I'd agree of having it behind a permission.

extendedclip:
The problem is /daddy will still return the no permission message.

BomBardyGamer:
This should've been PR'd just for the memes lol

PR #411, opened 2020-07-31
Dynamic Command Description, JSON Messages

Frcsty:
## Please read

### Type

- [] Internal change (Doesn't affect end-user).
- [ ] Other: __________ 

Dynamic command descriptions, fancified messages

### DIFF
--- a/src/main/java/me/clip/placeholderapi/util/ColorPalette.java
+++ b/src/main/java/me/clip/placeholderapi/util/ColorPalette.java
@@ -1,0 +1,42 @@
+package me.clip.placeholderapi.util;
+
+import net.md_5.bungee.api.ChatColor;
+import org.bukkit.Bukkit;
+
+public enum ColorPalette
+{
+
+    // MAIN
+    MAIN_BLUE("#1aaec4", ChatColor.AQUA),
+    MAIN_WHITE("#dae1e3", ChatColor.WHITE),
+    MAIN_GRAY("#909191", ChatColor.GRAY),
+
+    // ERRORS
+    ERROR_RED("#fa4732", ChatColor.DARK_RED),
+    ERROR_GRAY("#4d4d4d", ChatColor.GRAY),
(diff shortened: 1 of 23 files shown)


src/main/java/me/clip/placeholderapi/util/ColorPalette.java
```diff
@@ -0,0 +1,42 @@
...
+    private final String hex;
+    private final ChatColor color;
+
+    ColorPalette(final String hex, final ChatColor color) {
+        this.hex = hex;
+        this.color = color;
+    }
+
+    public ChatColor getColor() {
+        if (Bukkit.getServer().getClass().getPackage().getName().contains("16")) {
```
Andre601, line 35: Since we use our own version of JSONMessage could this be improved a lot (Especially for future versions) by using `ReflectionHelper.MAJOR_VER >= 16`

extendedclip: I agree with this

src/main/java/me/clip/placeholderapi/util/ColorPalette.java
```diff
@@ -0,0 +1,42 @@
...
+            return ChatColor.of(hex);
```
Andre601, line 36: This can have HEAVY issues when it would be used with the ITALIC or UNDERLINE enum field.
I don't say it will and I believe it won't be, but it could throw an IllegalArgumentException, because the `of` method makes a try-catch where it parses an integer from the provided String (which would fail in those cases.

You should add an `if (hex.equals("&o")) return ChatColor.ITALIC` (Same for Underline)

extendedclip:
There are a few things I want to nitpick with this pr so I don't really think its ready to be merged and released with 2.10.8. We will keep this open and work on it after that.

Andre601:
No longer worked on by Frosty

PR #420, opened 2020-08-04
Fix/419 set placeholders

### LINKED ISSUE #419: [Discussion] Do not use OfflinePlayer and enforce non-null online players
### Type
- [x] Change to code (External).  

### Description
Whoever made this is someone who doesn't understand api design
image 

My request is simple: enforce online players. 
Why? First, backwards compatibility. API design is the politics game and you can't go breaking it without solid argument. I don't see ANY solid argument for breaking the working PlaceholderAPI#setPlaceholders(Player, String) . 
Second, the majority of expansions and plugins who add placeholders require online player to function properly and changing that would break MASSIVE amount of plugins adding their placeholders which would result in children reviewing 1 star completely working plugins. Yes java will fallback to the offline player method for plugins only applying placeholders, that's not the problem here. The problem is the expansion support and plugins adding placeholderapi right now which compile against 2.10.8, which requires a ugly cast to make ide not scream at you. 

And now, a message to the one who made the change I have showed at the start (_credits to aikar and starlis/empirecraft for writing that short explanation_):
API's are tough to design. In public projects such as Bukkit, Spigot, Sponge etc, once an API is commited. **It's almost forever. You can't go breaking it without solid justification. This is the politics game.**

Andre601:
### Type

- [ ] Other: __________ 

### Description

Closes #419 

This undoes the deprecation of the `setPlaceholders` and `setBracketPlaceholders` methods that use a Player instance (With exception of the methods with the colorize param. Thos stay deprecated).

The Player instance is marked as `@NotNull` and in addition was the `@Nullable` annotation removed from the `OfflinePlayer` params to avoid issues.


MrIvanPlays:
`Can be null` is also an issue. Shouldn't be null

Andre601:
> `Can be null` is also an issue. Shouldn't be null

Not really. The main point of OfflinePlayer was and is that you can pass `null` as an argument.

Andre601:
Superseded by eeb9c8d5db64416b6fe80e5bf678d2050c9b41fe

Issue #413, opened 2020-08-02
Random stacktrace upon start


clrxbl:
### Issue
> What is the issue? Describe it like you would tell a friend.

Random stacktrace upon server start, plugin seems to function normally otherwise

```
[17:26:55 WARN]: [PlaceholderAPI] failed to download expansion information
java.util.concurrent.CompletionException: java.util.ConcurrentModificationException
at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273) ~[?:1.8.0_242]
at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280) ~[?:1.8.0_242]
at java.util.concurrent.CompletableFuture.uniApply(CompletableFuture.java:618) ~[?:1.8.0_242]
at java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:591) ~[?:1.8.0_242]
at java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:457) ~[?:1.8.0_242]
at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289) ~[?:1.8.0_242]
at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056) ~[?:1.8.0_242]
at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692) ~[?:1.8.0_242]
at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157) ~[?:1.8.0_242]
Caused by: java.util.ConcurrentModificationException
at java.util.HashMap$ValueSpliterator.tryAdvance(HashMap.java:1651) ~[?:1.8.0_242]
at java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:126) ~[?:1.8.0_242]
at java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:499) ~[?:1.8.0_242]
at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:486) ~[?:1.8.0_242]
at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472) ~[?:1.8.0_242]
at java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:152) ~[?:1.8.0_242]
[...]

PR #422, opened 2020-08-05
Attempt at fixing 413

MrIvanPlays:
### Type

### Description
This is my (miserable) attempt at fixing #413
These changes basically fix some potential threading issues and (probably) #413
Local tests went fine for me, but more tests are required.


extendedclip:
Looks good from the tests so lets see

PR #423, opened 2020-08-05
Refactorization of the TimeUtil utility class

Fefo6644:
### Type

- [ ] Other: __________ 

#### Refactored `TimeUtil` utility class
* Minimized `TimeUtil#getRemaining`.
* Fixed a bug regarding how days and hours were calculated in `TimeUtil#getTime`.
* Ditched the use of `int`s and use `long`s instead.
* Implemented `int`-taking methods so it won't break for the already existing uses through interfaces.
* Use `StringJoiner` instead of `StringBuilder` for simplicity's sake.

Closes N/A 


HelloZhangf:
Can you fix that? 
The time format for bug.
🤪

PR #436, opened 2020-08-11
Update to 1.16.2 + Inline bStats

darbyjack:
### Type

- [ ] Other: __________ 

### Description

Compiles the plugin under 1.16.2 and inlines + upgrades bStats.
This fixes issues with people needed bStats in their pom to compile PlaceholderAPI.

Closes N/A 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,9 +19,8 @@ repositories {
 
 dependencies {
     implementation "com.google.code.gson:gson:2.8.6"
-    implementation "org.bstats:bstats-bukkit:1.5"
 
-    compileOnly "org.spigotmc:spigot-api:1.16.1-R0.1-SNAPSHOT"
+    compileOnly "org.spigotmc:spigot-api:1.16.2-R0.1-SNAPSHOT"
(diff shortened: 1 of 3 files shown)


PiggyPiglet:
PlaceholderAPI compiles fine as is, not sure why inlining bstats was necessary.

build.gradle
```diff
@@ -47,7 +46,6 @@ tasks.withType(JavaCompile) {
 shadowJar {
     archiveClassifier.set("")
 
-    relocate "org.bstats", "me.clip.placeholderapi.metrics"
```
Andre601, line 50: This should stay relocated or it might conflict with any other plugin that uses bStats without relocation (or even with Spigot's bStats itself)

Andre601: Just now saw the Metrics clas.... WHY?
This is pointless and just too much to just fix a download issue that (so far) barely anyone had.

extendedclip:
I don't see a need to inline bStats as I have no issues on all 3 of my computers to compile papi in different environments.

PR #437, opened 2020-08-12
Adding Label commenter

Andre601:
### Type

- [x] Other: Issues/Pull requests 

### Description
This adds a GitHub Action, which allows us to send canned responses based on labels we (un)assign to issues and Pull requests.

For example could we send a Comment to an issue when the `Type: Issue (Expansion)` label is assigned, informing about the issue being caused by an Expansion.

Closes N/A 

### DIFF
--- a/.github/label-commenter-config.yml
+++ b/.github/label-commenter-config.yml
@@ -1,0 +1,57 @@
+labels:
+  - name: 'Type: Issue (Expansion)'
+    labeled:
+      issue:
+        body: |-
+          Hello @{{ sender.login }},
+          
+          The issue you encountered is caused by an Expansion and not PlaceholderAPI itself.
+          This issue-tracker is reserved for Bug reports and feature requests towards PlaceholderAPI.
+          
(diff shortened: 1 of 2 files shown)


Andre601:
I added a label-commenter-config.yml which currently has the following settings for the following labels:

**Type: Issue (Expansion)**
Comments on the issue about it being caused by an expansion and not PlaceholderAPI and that the issue should be reported to the Expansion's main issue tracker.

**Type: Duplicate**
Comments on the issue informing about the fact that another issue or Pull request exists, which is the exact same.

Closes the Issue/Pull request.

**Problem: More info required**
Tells the author to provide more information.
It will also comment on the issue when the label gets removed.

**Type: Invalid**
Tells the author that the issue is invalid which may be caused by it either not following any template, or by it reporting an issue that is in no way related/caused by PlaceholderAPI.

Closes the issue.

**Target: Wiki**
Informs the author that changes to the wiki are now handled on the repository and that they can submit their own changes through a PR.

Feedback would be highly appreciated.

PR #439, opened 2020-08-13
Added RocketPlaceholsers

Lorenzo0111:
### Type

- [ ] Other: __________ 

### Description

Added RocketPlaceholsers which provides their own placeholders. The placeholders are customizable from the config. 


Andre601:
When your plugin provides own placeholders should you also add it to the Placeholders page too.

Andre601:
The wiki should be updated shortly. Thanks for supporting PlaceholderAPI

PR #440, opened 2020-08-13
Fix PlaceholderAPI sending update notifications on Join despite it being disabled

Andre601:
### Type

- [ ] Other: __________ 

### Description
Right now does PAPI send an update notification to a joining player, even if "check_updates" is set to false. This is fixed with this PR.

Additionall was the Join listener moved from the UpdateChecker class to its own class in the listener folder (Makes more sense)

Closes N/A 

### DIFF
--- a/src/main/java/me/clip/placeholderapi/listeners/JoinListener.java
+++ b/src/main/java/me/clip/placeholderapi/listeners/JoinListener.java
@@ -1,0 +1,41 @@
+package me.clip.placeholderapi.listeners;
+
+import me.clip.placeholderapi.PlaceholderAPI;
+import me.clip.placeholderapi.PlaceholderAPIPlugin;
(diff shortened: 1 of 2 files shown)


extendedclip:
I don't think this is necessary. The current update checker only checks if enabled on startup and does not register the listener in the case it is disabled. There is nothing to fix as it is intentional to be that way. No need for dynamic update checker enable/disable. Either disable it and keep it that way or enable it and keep it that way.

Andre601:
You missunderstand.
The issue here is, that PAPI checks and informs about new versions EVEN when you have it disabled, which this PR fixes

extendedclip:
Incorrect....

onEnable

extendedclip:
the fetch method registers the listener ONLY if there is an update available and that will never be called on start if check updates is set to false.

Issue #444, opened 2020-08-18
[Suggestion] Add event for when all expansions have been loaded


Andre601:
### Type

- [x] New function for PlaceholderAPI.  
A new function that developers could use.
- [ ] Change to code (Internal).  
Changes to code that won't affect the end-user.
- [ ] Change to code (External).  
Changes to code that will affect the end-user (breaks stuff).
- [ ] Other: __________ 

### Description

It would be beneficial when PlaceholderAPI has some event that would be fired once all its expansions have been loaded.
This would be useful for plugins that depend on PlaceholderAPI and on the fact that all expansions have been loaded.

PR #460, opened 2020-08-31
Added event when all expansions are loaded for developers.

ignissak:
Event `ExpansionsLoadedEvent` is called on the end of `LocalExpansionManager#registerAll()`.
Resolves https://github.com/PlaceholderAPI/PlaceholderAPI/issues/444

### DIFF
--- a/src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
+++ b/src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
@@ -322,8 +322,7 @@ public final class LocalExpansionManager implements Listener {
           registered == 0 ? "&6No expansions were registered!"
               : registered + "&a placeholder hooks successfully registered!");
 
-      final ExpansionsLoadedEvent event = new ExpansionsLoadedEvent();
-      Bukkit.getPluginManager().callEvent(event);
+      Bukkit.getPluginManager().callEvent(new ExpansionsLoadedEvent());
     });
   }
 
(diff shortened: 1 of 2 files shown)


src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
```diff
@@ -320,6 +321,9 @@ private void registerAll(@NotNull final CommandSender sender) {
       Msg.msg(sender,
           registered == 0 ? "&6No expansions were registered!"
               : registered + "&a placeholder hooks successfully registered!");
+
+      final ExpansionsLoadedEvent event = new ExpansionsLoadedEvent();
```
Andre601, line 325: Why do it like that and not just `Bukkit.getPluginManager().callEvent(new ExpansionsLoadedEvent())`?

ignissak: You are right, I am gonna change it.

ignissak: Changed.

Andre601 (approved):

darbyjack (approved):

NEZNAMY:
Can there also be an API method to check if expansions are loaded in case we "miss" the event? Such as when our plugin is reloaded and we need to check if expansions are loaded.

Andre601:
> Can there also be an API method to check if expansions are loaded in case we "miss" the event? Such as when our plugin is reloaded and we need to check if expansions are loaded.

To check for a specific expansion:
https://github.com/PlaceholderAPI/PlaceholderAPI/blob/master/src/main/java/me/clip/placeholderapi/PlaceholderAPI.java#L204-L213

To get all currently loaded ones:
https://github.com/PlaceholderAPI/PlaceholderAPI/blob/master/src/main/java/me/clip/placeholderapi/PlaceholderAPI.java#L215-L224

PR #485, opened 2020-09-24
Remove Gson

### LINKED ISSUE #477: PAPI gutting other plugins' and its own bStats
### Issue
PlaceholderAPI includes, and relocates, its own Gson. PAPI also includes, and relocates, bstats-bukkit.

bStats uses Gson to collect plugin data. It uses reflection to invoke other metrics instances' `getPluginData` methods then instanceof checks these based on `JsonObject`.

https://github.com/Bastian/bStats-Metrics/blob/ed06bb68cc180e108546e686e7b5124cde20cd5d/bstats-bukkit/src/main/java/org/bstats/bukkit/Metrics.java#L282

However, due to PAPI's relocation, JsonObject from PAPI is no longer JsonObject from the server's unrelocated Gson.

https://github.com/PlaceholderAPI/PlaceholderAPI/blob/a74f78bd732c0094152be92f265f822efd77edd7/build.gradle#L50

The way bStats works, only 1 metrics instance submits data and draws data from the other instances. If PAPI becomes the main instance, it does not pick up other instances' data since the instanceof check fails. If another plugin becomes the main instance, it does not pick up PAPI's data for the same reason.

### Expected behaviour

PAPI's bStats, when submitting data, should recognise other plugins' bStats data. Other plugins bStats, when submitting data, should recognise PAPI's data.

### Actual behaviour
PAPI ignores other plugins' data because the instanceof check in bStats fails. Likewise for other plugins ignoring PAPI's data.

### How to reproduce
Copy and paste metrics class to own plugin and add debug statements to metrics class's `submitData` method. Run the debug plugin alongside PAPI and have it call the `submitData` method. You will see that PAPI's data is ignored, and you will also see that, if PAPI submits metrics, the other plugin's data is ignored.

Result:
http://prntscr.com/uhb6v4
http://prntscr.com/uhb6t9

### `/papi dump` output
I believe I have enough documentation and reproduction steps, but if you really need me to get the output of this command, I still can.

Andre601:
### Type

- [ ] Other: __________ 

### Description

I made this PR now to finally get this rolling.
As described in #477 does PAPI have a chance of messing up bStats which may or may not break stats of all the other plugins that use bStats.
From what I gathered in the issue is the problem the relocated gson and the only reason gson is there is because of 1.7 support.

I say fuck 1.7! That version doesn't have any real reason to be supported on our end and finding a complicated solution (like making a separate jar... why?) just to still support it is imo not good at all.

This is a draft to keep a sort of discussion running so that we can (finally) decide on a solution. If there is a better solution without making PAPI too overcomplicated in terms of maintenance should this be mentioned.
I may open this for merge once we finally go to a proper point but my opinion is to finally drop 1.7 support. It's not worth the trouble in any way.

Closes #477 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,6 @@ repositories {
 }
 
 dependencies {
-    implementation "com.google.code.gson:gson:2.8.6"
     implementation "org.bstats:bstats-bukkit:1.5"
 
     compileOnly "org.spigotmc:spigot-api:1.16.2-R0.1-SNAPSHOT"
@@ -48,7 +47,6 @@ shadowJar {
     archiveClassifier.set("")
 
     relocate "org.bstats", "me.clip.placeholderapi.metrics"
-    relocate "com.google.gson", "me.clip.placeholderapi.libs.gson"
 }
 
 license {
--- a/src/main/java/me/clip/placeholderapi/commands/impl/local/CommandDump.java
+++ b/src/main/java/me/clip/placeholderapi/commands/impl/local/CommandDump.java
@@ -21,7 +21,17 @@
 package me.clip.placeholderapi.commands.impl.local;
 
 import com.google.common.io.CharStreams;
-import com.google.gson.JsonParser;
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import me.clip.placeholderapi.PlaceholderAPIPlugin;
+import me.clip.placeholderapi.commands.PlaceholderCommand;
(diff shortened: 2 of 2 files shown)


Andre601:
@darbyjack @extendedclip @PiggyPiglet Any input on this? Should we just drop 1.7 support? I think that really is the best option here.

PiggyPiglet (approved):

PR #517, opened 2020-12-27
Fixing issues for logging when Expansion doesn't contain any classes.

### LINKED ISSUE #515: Error when registering expansions
### Issue
Expansions are not successfully registered after downloading using `papi ecloud download <expansion>`.

### Expected behaviour
PAPI should download, prompt to reload, and register the expansion after reloading.

### Actual behaviour
PAPI downloads the expansion, prompts to reload, but throws null pointer exceptions when reloading.
https://paste.helpch.at/redubilofa.bash

### How to reproduce

1. Download expansion using `papi ecloud download server`
2. `Successfully downloaded expansion Server [2.4.0] to file: Expansion-server.jar
Make sure to type /papi reload to enable your new expansion!`
3. Reload using `papi reload`

### `/papi dump` output
When I've been downloading expansions all names have been "Expansion-[lowercase plugin name]". Reloading PAPI does not register the newly downloaded expansions automatically. After looking at other issues and dumps, I realized that my expansion for server was called "Expansion-server.jar" in my dump, but "Expansion-Server.jar" in another user's dump. I redownloaded server from the ecloud and console still reported
`Successfully downloaded expansion Server [2.4.0] to file: Expansion-server.jar
Make sure to type /papi reload to enable your new expansion!`
First Dump https://paste.helpch.at/obapozepuf
After renaming the downloaded file to "Expansion-Server.jar" and then reloading PAPI, the expansion was successfully registered and recorded in a subsequent dump. https://paste.helpch.at/wesawaseto

Is there a list of the names or syntax PAPI looks for? `papi ecloud list all`? Some don't register using their ecloud name either. Example: "mcMMO" as seen in `papi ecloud list all` isn't registered when named "Expansion-mcmmo.jar" or "Expansion-mcMMO.jar" but is registered when named "Expansion-McMMO.jar".
I cannot get ListPlayers and SpeedPerSec to register. All of the others I have renamed and were successfully registered.
https://paste.helpch.at/olopujunoy
If I remove ListPlayers and SpeedPerSec the above null pointer exception also disappears.
https://paste.helpch.at/ruhefidomo
I'd really like to be able to just download, reload, and move on to implementation. The plugin should get the placeholders for me without needing me to double check the files are named correctly.

### LINKED ISSUE #516: [Proposal] Improve logging of errors and similar
### Type

- [ ] New function for PlaceholderAPI.  
A new function that developers could use.
- [x] Change to code (Internal).  
Changes to code that won't affect the end-user.
- [ ] Change to code (External).  
Changes to code that will affect the end-user (breaks stuff).
- [ ] Other: __________ 

### Description

The current logging seems a bit odd to me.
For example is here an example of the register method in the LocalExpansionManager:
https://github.com/PlaceholderAPI/PlaceholderAPI/blob/4074192c0bea102b106d6cb63d9ed71cad53149a/src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java#L170-L172

Why use ex.getMessage()? If it's an exception then it should be thrown for clearance as to where it happened.
And instead of "expansion class xxx is outdated" have something like "Exception while registering expansion xxx (Is it outdated?)"

Andre601:
### Type

- [ ] Other: __________ 

### Description

Closes #515
Closes #516

When an expansion doesn't contain any class files (Being it malformed download or broken build) will PlaceholderAPI throw an exception when it attempts to log the failed loading of an expansion, essentially breaking the logging. Please see #515 for full context.

This Pull request is NOT finished yet as we still need to find a solution to fix this issue while still logging the failed load of an expansion.

In addition have some of the lines been updated to not provide missinformation (Often such errors happen due to a missing dependency and not an outdated expansion)

### DIFF
--- a/src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
+++ b/src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
@@ -22,27 +22,11 @@ package me.clip.placeholderapi.expansion.manager;
 
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
-import java.io.File;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.locks.ReentrantLock;
-import java.util.logging.Level;
 import me.clip.placeholderapi.PlaceholderAPIPlugin;
 import me.clip.placeholderapi.events.ExpansionRegisterEvent;
 import me.clip.placeholderapi.events.ExpansionUnregisterEvent;
 import me.clip.placeholderapi.events.ExpansionsLoadedEvent;
-import me.clip.placeholderapi.expansion.Cacheable;
-import me.clip.placeholderapi.expansion.Cleanable;
-import me.clip.placeholderapi.expansion.Configurable;
-import me.clip.placeholderapi.expansion.PlaceholderExpansion;
-import me.clip.placeholderapi.expansion.Taskable;
-import me.clip.placeholderapi.expansion.VersionSpecific;
+import me.clip.placeholderapi.expansion.*;
 import me.clip.placeholderapi.expansion.cloud.CloudExpansion;
 import me.clip.placeholderapi.util.FileUtil;
 import me.clip.placeholderapi.util.Futures;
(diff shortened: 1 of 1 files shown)


PiggyPiglet:
In the non null filter, we need to log an error saying which was null, so the user can remove that expansion or fix it.

Andre601:
> In the non null filter, we need to log an error saying which was null, so the user can remove that expansion or fix it.

Which one do you mean with "non null filter"? The `filter(Objects::nonNull)`?

PiggyPiglet:
> > In the non null filter, we need to log an error saying which was null, so the user can remove that expansion or fix it.
> 
> Which one do you mean with "non null filter"? The `filter(Objects::nonNull)`?

yeah

Andre601:
> > > In the non null filter, we need to log an error saying which was null, so the user can remove that expansion or fix it.
> > 
> > 
> > Which one do you mean with "non null filter"? The `filter(Objects::nonNull)`?
> 
> yeah L#328 LocalExpansionManager

The bigger issue here is, that we can't even log properly as a null object such as the file's name is enough to break things.
See Line 364 (Currently commented out)
If there is an alternative method to get the name should it be mentioned.

Andre601:
Additionally does the Objects::nonNull check not work either as the stream there is only looking for the jar files to exist (which they do) but doesn't do checks for if the file is actually valid and not corrupted.

PiggyPiglet (approved):

PR #547, opened 2021-01-24
Improve bStats metrics

Andre601:
### Type

- [ ] Other: __________ 

### Description

Closes N/A 

This PR changes a few things.
It first of all changes the `Version` class to have a nested `Type` Enum which is used in the `static` of `PlaceholderAPIPlugin` to determine the actual server type used.

The code to do this has been borrowed from the Server-Expansion and currently supports Purpur, Tuinity, PaperMC and Spigot itself.

Next does `using_spigot` now return 3 different responses:

- `yes (fork)` when it is a Spigot fork (Currently PaperMC, Tuinity and Purpur)
- `yes` for Spigot itself
- `no` for any other cases

All forks are currently treated as `isSpigot = true`, so if that should be changed, let me know.

The original `Version(String, boolean)` Constructor has been deprecated in favour of the `Version(String, Version.Type)` Constructor, but still works, with the class resulting in having `Version(version, Type.SPIGOT)` set when server is a Spigot server and `Version(version, Type.UNKNOWN)` when it is not.

**Important**
A new SimplePie Chart has been added. The identifier is `server_brand` and it returns the Server Brand name, so a new Chart would need to be added to bStats to have these metrics working.

### DIFF
--- a/src/main/java/me/clip/placeholderapi/PlaceholderAPIPlugin.java
+++ b/src/main/java/me/clip/placeholderapi/PlaceholderAPIPlugin.java
@@ -239,9 +239,6 @@ public final class PlaceholderAPIPlugin extends JavaPlugin {
           
           return map;
         }));
-    
-    metrics.addCustomChart(
-        new Metrics.SimplePie("server_brand", () -> getServerVersion().getName()));
 
     metrics.addCustomChart(new Metrics.AdvancedPie("expansions_used", () -> {
       final Map<String, Integer> values = new HashMap<>();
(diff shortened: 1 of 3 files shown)


Andre601:
I decided to now have the server brand as its own SimplePie Chart with the identifier `server_brand`, so if this gets merged should the Metrics on bStats receive a new chart with that identifier.

Additionally does `using_spigot` now return `yes (fork)` when the Server brand is a fork of Spigot (Currently everything except Spigot and Unknown). It will otherwise either return `yes` for normal Spigot Server or `no` in any other case.

PiggyPiglet (approved):

Andre601:
This won't be added in the near future, so I just close this.

PR #575, opened 2021-02-17
First attempt at adding Hex support.

### LINKED ISSUE #573: Update RGB to parse &#rrggbb in addition to &x&r&r&g&g&b&b
### Type

- [x] New function for PlaceholderAPI.  
A new function that developers could use.
- [x] Change to code (Internal).  
Changes to code that won't affect the end-user.
- [ ] Change to code (External).  
Changes to code that will affect the end-user (breaks stuff).
- [ ] Other: __________ 

### Description

Suggestion to include the ability for PAPI to parse RGB color codes in the &#rrggbb format as it is becoming increasingly more popular with other plugins. Currently users have to create multiple versions of their RGB colors to parse in PAPI and parse in other plugins. If PAPI was able to parse both formats, this would solve much headache and allow a single format to work across all plugins which support PAPI and/or either of these formats for parsing.

image
image

The value for displayname is &#f426caO&#ff4176w&#ff9230n&#e2d128e&#93ff85r&f
The value for displayname-papi is &x&f&4&2&6&c&aO&x&f&f&4&1&7&6w&x&f&f&9&2&3&0n&x&e&2&d&1&2&8e&x&9&3&f&f&8&5r&f

Andre601:
### Type

- [ ] Other: __________ 

### Description

Closes #573

Adds `&#<hex>` color format to PAPI by using the BungeeCord ChatColor class.

The CharReplacer has been updated to acknowledge the new color syntax and to also only translate to valid hex colors. Right now tho does it remove the `&x` from the start, but I'm not sure how to correct this or IF it should be corrected at all.

JSONMessage class also was updated to support the ChatColor class of BungeeCord and from my basic testing did it work.

Here's the latest Jar for you to try out, if you like.
https://cdn.discordapp.com/attachments/584040473186926623/811647295052447814/PlaceholderAPI-2.10.10-DEV-null.jar

### DIFF
--- a/src/main/java/me/clip/placeholderapi/commands/impl/cloud/CommandECloudExpansionList.java
+++ b/src/main/java/me/clip/placeholderapi/commands/impl/cloud/CommandECloudExpansionList.java
@@ -46,7 +46,7 @@ import me.clip.placeholderapi.expansion.cloud.CloudExpansion;
 import me.clip.placeholderapi.libs.JSONMessage;
 import me.clip.placeholderapi.util.Format;
 import me.clip.placeholderapi.util.Msg;
-import org.bukkit.ChatColor;
+import net.md_5.bungee.api.ChatColor;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
 import org.jetbrains.annotations.NotNull;
--- a/src/main/java/me/clip/placeholderapi/libs/JSONMessage.java
+++ b/src/main/java/me/clip/placeholderapi/libs/JSONMessage.java
@@ -27,8 +27,9 @@ import java.util.List;
 import java.util.Objects;
 import java.util.UUID;
 import java.util.Vector;
+
+import net.md_5.bungee.api.ChatColor;
 import org.bukkit.Bukkit;
-import org.bukkit.ChatColor;
 import org.bukkit.entity.Player;
 
 /**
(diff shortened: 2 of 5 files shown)


Andre601:
This seems to be a bigger issue than expected. Please see the linked issue for context about what isn't working.

Any input on how this could be solved is appreciated, as I myself have no clue what else could be done to fix this problem.

MrIvanPlays:
The problem here is the multi version support. I think we need an adapter library. Like ditch all shit PAPI has about colors and messages and redo them from ground so it supports all the versions PAPI intends to support. Cuz like it is using the 1.16 version of the BungeeCord Chat API, which has a lot of stuff different compared to the 1.15 one.

Andre601:
> The problem here is the multi version support. I think we need an adapter library. Like ditch all shit PAPI has about colors and messages and redo them from ground so it supports all the versions PAPI intends to support. Cuz like it is using the 1.16 version of the BungeeCord Chat API, which has a lot of stuff different compared to the 1.15 one.

Yeah, that would indeed be an issue.
I, unfortunately, don't have enough knowledge about PAPIs Char replacing system (Thanks Sxtana for that) to actually come up with an idea on how to solve this in an elegant way.

Andre601:
I close this as I don't feel like PlaceholderAPI should do any Color formatting for things such as `setPlaceholders`.

It should be the job and priority of plugins that manage wherever the placeholder is displayed (Chat, Tab list, Scoreboard, ...), or the expansion itself.

PR #593, opened 2021-03-21
Update LuckPerms placeholders

lucko:
### Type

- [ ] Other: __________ 

### Description

Updates the documentation to correctly point to the LuckPerms wiki for a full list of placeholders.

I considered just updating the list to include the new/altered placeholders, however, I think:
* it makes sense to have a single source - only one place to update in the future
* the descriptions (missing from the docs here) are quite important for end-users to see and understand, as it is not immediately obvious what some of the placeholders do just by their name.

Thanks :)


Andre601 (approved):
Thank you for this update

PR #601, opened 2021-03-26
Report command

Andre601:
### Type

- [ ] Other: __________ 

### Description

Closes N/A 

Adds the `report` command to PlaceholderAPI.

This command will do a few things: It will create a dump similar to that of `/papi dump` and also upload the latest.log content to the paste site.
Those URLs are then used in a pre-made URL that would open a GitHub issue with the Papi dump and Console field being filled out.

Latest working URL that was generated by this command:
https://github.com/PlaceholderAPI/PlaceholderAPI/issues/new?labels=Type%3A+Issue+%28Unconfirmed%29&template=bug_report.yml&dump=https%3A%2F%2Fpaste.helpch.at%2Fvumavoqihe&console=https%3A%2F%2Fpaste.helpch.at%2Falekuxikup

It should perhaps be considered to use a dedicated URL for this instead (i.e. `https://report.helpch.at?dump=<url>&console=<url>`) to then forward to the above URL.
The fields `dump` and `console` are from PAPI while all the others are from selecting an issue, so they're mandatory to not break the existing Issue system.


Andre601:
After looking at this again do I see no real benefit in this command to begin with. Closing it.

PR #612, opened 2021-04-13
Remove color formatting from placeholder parser

Andre601:
### Type

- [ ] Other: __________ 

### Description
This was made with a huge help from Crypto Morin.

This PR **removes** the color code parsing from PlaceholderAPI's CharsReplacer.
It shouldn't be the Job of PlaceholderAPI to parse the colors since that only causes unwanted extra bloat and not needed work to maintain, especially with Hex colors in 1.16.

PlaceholderAPI should focus on parsing the placeholders themself. If the dev of an expansion wants translated colors should they do it themself.

Some other small changes:
- Removed invalid `@NotNull` annotation from `getBoolean` method since it can't be applied to primitives.
- Made Version a static class
- Added some checks to `findExpansionsOnDisk` for the case that the returned `File` array is null.

This is obviously a breaking change so a merge should only be made after it has been discussed AND when PAPI is ready for the next major release.

### DIFF
--- a/src/main/java/me/clip/placeholderapi/events/ExpansionsLoadedEvent.java
+++ b/src/main/java/me/clip/placeholderapi/events/ExpansionsLoadedEvent.java
@@ -26,7 +26,7 @@ import org.bukkit.event.HandlerList;
 import org.jetbrains.annotations.NotNull;
 
 /**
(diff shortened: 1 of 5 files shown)


Andre601:
Will re-open this against the master branch

PR #614, opened 2021-04-14
Update PlaceholderExpansion.md

UntouchedOdin0:
The org.bukkit. wasn't necessary, so I'm proposing for it to be removed from here.

### Type

- [ ] Other: __________ 

### Description

Closes N/A 


Andre601:
The exclusion of it is also not needed either. And it's often always better to have the full Path instead of the final class to prevent possible conflicts.

PR #630, opened 2021-05-12
Has change and add descriptions

broken1arrow:
Try other way and see if this work better to pull uppdates.

### Type

- [ x] Wiki (Changes towards the [Wiki]).
- [ ] Other: __________ 

### Description

### DIFF
--- a/wiki/Placeholders.md
+++ b/wiki/Placeholders.md
@@ -20,21 +20,21 @@ If the command itself isn't there and `NO DOWNLOAD COMMAND` instead is shown, th
     - **[Animations](#animations)**
     - **[Armor](#armor)**
     - **[ASCII](#ascii)**
-  
+
   - ### **B**
     - **[BungeeCord](#bungeecord)**
-  
+
   - ### **C**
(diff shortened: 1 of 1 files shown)


broken1arrow:
Seams

Andre601:
Lots of pointless changes (removal of spaces) that aren't needed.

Also, to mention something different real quick: Your expansion has a lot of flaws in it. From bad coding practices to overall design.
For once is reloading the config every time a placeholders is requested not a good idea as this can happen at least 20 times a second(!)

broken1arrow:
> Lots of pointless changes (removal of spaces) that aren't needed.
> 
> Also, to mention something different real quick: Your expansion has a lot of flaws in it. From bad coding practices to overall design.
> For once is reloading the config every time a placeholders is requested not a good idea as this can happen at least 20 times a second(!)

yea was my edit program som mess it up.

I try to clean the code up for new update, is not perfect, but are better an first version I realize. Will fix up the code on github later on when fix more on the code. If you has more inputs I´m only happy to fix the issues.

Will move config loading to better place (I have to think out beter metod to load the yml).

PR #656, opened 2021-06-30
Small improvements (cleanup annoying error messsage, fix version checker)

Starmism:
### Type
- [ ] Other: __________ 

### Description
Fixes an issue with the version checker not working correctly for `10`, since it parses trailing zeroes as non-existent.
Clarifies the source of an error where the `getAuthor()` method of an expansion returns null. It now shows which expansion it came from.
Cleans up and updates the build.gradle.

Closes N/A


src/main/java/me/clip/placeholderapi/expansion/PlaceholderExpansion.java
```diff
@@ -46,31 +45,27 @@
    *
    * @return placeholder identifier that is associated with this expansion
    */
-  @NotNull
```
Andre601, line 49: Why do you remove the Annotations?
The Strings are required to be non-null and the annotation helps with this in IDEs

src/main/java/me/clip/placeholderapi/updatechecker/UpdateChecker.java
```diff
@@ -90,17 +91,22 @@ private boolean spigotIsNewer() {
...
-    String spV = toReadable(spigotVersion);
-    return plV.compareTo(spV) < 0;
+    int[] plV = toReadable(pluginVersion);
+    int[] spV = toReadable(spigotVersion);
+
+    if (plV[0] < spV[0]) {
+      return true;
+    } else if ((plV[1] < spV[1])) {
+      return true;
+    } else return plV[2] < spV[2];
```
Andre601, line 101:
```suggestion
    } else {
      return plV[2] < spV[2];
    }
```
Just to keep it consistent.

src/main/java/me/clip/placeholderapi/updatechecker/UpdateChecker.java
```diff
@@ -90,17 +91,22 @@ private boolean spigotIsNewer() {
...
-  private String toReadable(String version) {
-    if (version.contains("-DEV-")) {
-      version = version.split("-DEV-")[0];
+  private int[] toReadable(String version) {
+    if (version.contains("-DEV")) {
+      version = version.split("-DEV")[0];
     }
 
-    return version.replaceAll("\\.", "");
+    return Arrays.stream(version.split("\\.")).mapToInt(Integer::parseInt).toArray();
```
Andre601, line 109: And what if this will for whatever reason be not a Number?
This would throw an exception without it being handled at all.

Andre601 (changes requested):

src/main/java/me/clip/placeholderapi/expansion/PlaceholderExpansion.java
Starmism: The whole problem stems from them being null anyways, despite being declared as NotNull. See [pastebin error](https://pastebin.com/B2EL8PEW)

src/main/java/me/clip/placeholderapi/updatechecker/UpdateChecker.java
Starmism: I doubt anyone will ever set the PAPI version number to a non-SemVer version. If they do, this error should be thrown, I think.

src/main/java/me/clip/placeholderapi/expansion/PlaceholderExpansion.java
Andre601: Still no need to remove the annotations, since they never caused errors so far.
At most add stuff to check for null and replace with default value or cancel expansion registration.
Removal of the snnotations does nothing here.

Starmism: True I'll give you that.

Starmism:
Closed in favor of #657.

PR #688, opened 2021-08-04
Update PlaceholderAPIConfig.java

SashaSemenishchev:
### Type

- [ ] Other: __________ 

### Description

Added stringNull() method to PlacegolderAPIConfig. Represents null value.

Closes N/A 


Andre601:
This is not even a config option...
https://github.com/PlaceholderAPI/PlaceholderAPI/blob/development/src/main/resources/config.yml

PR #759, opened 2021-12-03
Removed Deprecated Methods

darbyjack:
### Type

- [ ] Other: __________ 

### Description

In version `2.10.9`, we marked a bunch of older API methods as deprecated and to be removed in version `2.10.11`. The time is coming soon to release this new version so the methods marked for removal should be removed. Merge this last before the `2.11.0` release.

Closes N/A 


Andre601:
Uh.... Some are marked to get removed in `2.11.0` while you mention `2.10.11`, so which one is it now?

darbyjack:
Yup, that's a typo on my end. The next version *should* be 2.11.0 and then removed normally.

Andre601:
How's the status on this PR?
Anything else to update/remove?

darbyjack:
Going to hold back on this. Decided it was best to not take the chance to break stuff before PAPI3.

Andre601:
\*sigh* so you do the bukkit aproach here...
Just remove it! If people use it it's their fault. They had plenty of time. We should just release 2.11.0

And be honest: Do you REALLY believe PAPI3 will enter *any* alpha or beta state in the next year, given its current state?
I state the facts here.

emilyy-dev:
https://semver.org/

Andre601:
> https://semver.org/

Semver isn't everything.
While we can follow it, is it not required. Not to mention did we ANNOUNCE through the annotations used that the stuff WILL be removed in 2.11.0, so now changing that would also not be nice.

If we announce something, no matter in what way, should we stick with it.

emilyy-dev:
> > https://semver.org/
> 
> Semver isn't everything. While we can follow it, is it not required. Not to mention did we ANNOUNCE through the annotations used that the stuff WILL be removed in 2.11.0, so now changing that would also not be nice.
> 
> If we announce something, no matter in what way, should we stick with it.

I disagree, semver is a globally identifiable way of defining breaking changes, additions and patches/bug fixes that guarantees if my software changes a library version from X.Y.Z to X.W.V there will be no ABI breakages, regardless of context. A proper versioning system should always be put before some silly annotation pinky promise "no matter in what way, should we stick with it".

PR #800, opened 2022-02-05
Add --null argument for Parse command

### LINKED ISSUE #798: Suggestion: /papi parse null
### Confirmation

- [X] I checked the Issues and Pull request tab for any existing issues or PRs.
- [X] My Feature Request is for **PlaceholderAPI** and not any expansion or other plugin.

### Type

New API feature

### Description

There's currently a `/papi parse me %someplaceholder%`, alongside `/papi parse (playername) %someplaceholder%`
I suggest the addition of `/papi parse null %someplaceholder%` (or some other syntax to the same effect).
This would explicitly pass `null` as the player value to the placeholder, allowing the testing of placeholders in a global context.

This would help a lot when dealing with issues relating to placeholders that don't function when the player is missing.

In particular the main reason I want this myself is that I run support for the Citizens plugin, which often parses placeholders (for things like NPC names and skins) that apply in a global context, with a `null` player (these parts don't function on a per-player basis).
We occasionally get issue reports of placeholders 'just not working' on NPCs, and having a built-in command to parse explicitly with a null player would allow us to instantly distinguish something wrong on Citizens' end vs something wrong in the placeholder's code (there's a bunch of global placeholders out there that mistakenly return null when the player is null, despite not needing a player reference).

### Code Example

_No response_

### Jar file

_No response_

### Additional Information

_No response_

Andre601:
### Type

- [ ] Other: __________ 

### Description

This PR adds a `--null` command argument to the parse command, which can be used instead of the player name.
Providing this argument would parse the placeholders against a null player, which can be useful for Placeholders that do not require a online player to be used.

This PR also updates all `toLowerCase()` to `toLowerCase(Locale.ROOT)` to avoid possible issues where characters will be changed to non UTF-8 characters or smth (Idk the details but it's a good change for sure.)

Closes #798 


iGabyTM (approved):

HexedHero (approved):
Been using this and works fine

PR #801, opened 2022-02-05
Improve expansion sorting in /papi dump

Andre601:
### Type

- [ ] Other: __________ 

### Description

This changes the way expansions are sorted in the `/papi dump` command.

Previously did PlaceholderAPI sort the expansions by their Identifier and then by author.
Now it should sort them by identifier AND also by the author.

Idk... Nag @iGabyTM about this.

Closes N/A 

### DIFF
--- a/src/main/java/me/clip/placeholderapi/commands/impl/local/CommandDump.java
+++ b/src/main/java/me/clip/placeholderapi/commands/impl/local/CommandDump.java
@@ -106,7 +106,6 @@ public final class CommandDump extends PlaceholderCommand {
         }
 
         try (final InputStream stream = connection.getInputStream()) {
-          //noinspection UnstableApiUsage
           final String json = CharStreams.toString(new InputStreamReader(stream, StandardCharsets.UTF_8));
           return gson.fromJson(json, JsonObject.class).get("key").getAsString();
         }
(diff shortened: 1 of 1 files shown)


iGabyTM:
I don't see the point of having another sorting by author's name, isn't like there will be multiple expansions with the same ID or whatever. If you ask me, I would sort them only by identifier.

Andre601:
> I don't see the point of having another sorting by author's name, isn't like there will be multiple expansions with the same ID or whatever. If you ask me, I would sort them only by identifier.

The idea is to have the expansions grouped by the author, while in alphabetical order.
Would IMO be easier for debugging as all expansions are grouped by their author.

src/main/java/me/clip/placeholderapi/commands/impl/local/CommandDump.java
```diff
@@ -167,10 +168,15 @@ private String makeDump(@NotNull final PlaceholderAPIPlugin plugin) {
         .getExpansionsFolder()
         .list((dir, name) -> name.toLowerCase().endsWith(".jar"));
 
-    for (final String jar : jars) {
-      builder.append("  ")
-          .append(jar)
-          .append('\n');
+
+    if (jars == null) {
+      builder.append("  WARN: Jars array was empty!");
```
BlitzOffline, line 173: Shouldn't this say something more user friendly like `There were no JARs found in the extensions director`. From what I'm understanding this is what shows under `Expansion Directory:` a big warning that won't make sense for most people seems like the wrong approach.

Andre601: Since this would be printed when the array is null - which would be a rare error - would it perhaps be better to have it changed to `WARN: Could not collect Jar files`

BlitzOffline: Sure. I guess that does make sense. and it is more "user friendly"

darbyjack (approved):

PR #828, opened 2022-05-04
Add queue system for Expansion registration

Andre601:
### Type

- [ ] Other: __________ 

### Description

This is my attempt at creating a "queue" system for the expansion registration.

PlaceholderAPI right now has two distinct phases:

- Register expansions from plugins when their `register()` method is called
- Register expansions downloaded from the eCloud once server has completed its startup

This feels a bit weird and may even have weird results, if a plugin registers an expansion early on before anything else is loaded properly.
With this queue system, expansions should now be added to an internal queue and once the `registerAll()` method is called will PlaceholderAPI work its way through this queue.
Any future registration should then be registered directly.

I'll be honest here: Not the biggest fan of the way I implemented this, but I can't think of any better solutions, so please don't hesitate to suggest some alternative ways here.

From what I can see are there no breaking changes for the end-user as methods removed/changed where either declared private or annotated as Internal, so it would be their fault if stuff breaks for them.

### DIFF
--- a/src/main/java/me/clip/placeholderapi/PlaceholderAPIPlugin.java
+++ b/src/main/java/me/clip/placeholderapi/PlaceholderAPIPlugin.java
@@ -53,7 +53,6 @@ public final class PlaceholderAPIPlugin extends JavaPlugin {
   @NotNull
   private static final Version VERSION;
   private static PlaceholderAPIPlugin instance;
-  private boolean loaded = false;
 
   static {
     final String version = Bukkit.getServer().getClass().getPackage().getName().split("\\.")[3];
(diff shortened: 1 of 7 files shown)


Andre601:
I made this graph to hopefully explain how I imagine the new loading system to work:

```mermaid
flowchart TB
  
  plugin[ Plugin ]
  done[ Done ]

  queue[( Queue )]

  ready{{ PlaceholderAPI ready? }}
  empty{{ Queue empty? }}

  add> Add to Queue ]
  register> Register Expansions ]

  subgraph q [Queue]
    addJar> Add Expansion Jars to Queue ]
    mark> Mark PlaceholderAPI ready ]
    
    addJar --> mark
    mark --> register
    register --> empty
  end

  plugin -- "PlaceholderExpansion#register()" --> ready
  
  ready -- No --> add
  ready -- Yes --> register

  add --> queue

  queue -- ServerLoadEvent --> q

  empty -- Yes --> done
  empty -- No --> register
```

PR #829, opened 2022-05-05
Updates and softdepends

JT122406:
### Type

- [x] Other: Addition of Softdepends

### Description

Addition of Softdepends because log says it loads classes from these classes
Closes N/A 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,11 +23,11 @@ repositories {
 }
 
 dependencies {
-    implementation "org.bstats:bstats-bukkit:2.2.1"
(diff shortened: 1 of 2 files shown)


BlitzOffline:
The logs say that because the expansions themselves (valut, luckperms) load classes from them. So those warnings don't show if you don't have those expansions installed. If we go this way, PAPI will end up having every plugin in existence as a soft depend. which might end up being a problem since many plugins soft depend or even depend on papi.

JT122406:
ah that makes sense my bad

PR #773, opened 2021-12-25
Update Contributing file

Andre601:
### Type

- [ ] Internal change (Doesn't affect end-user).
- [ ] External change (Does affect end-user).
- [ ] Wiki (Changes towards the [Wiki]).
- [x] Other: CONTRIBUTING.md 

### Description

Since we apparently go away from the `development` branch and instead now only have the `master` branch did I quickly update the `CONTRIBUTING.md` file to reflect that change (And to hopefully make it a bit more understandable in some way).

Closes N/A 

[Wiki]: https://github.com/PlaceholderAPI/PlaceholderAPI/wiki

PR #407, opened 2020-07-31
Update Contributing file

Andre601:
## Please read
Please make sure you checked the following:
- You checked the [Pull requests] page for any upcoming changes.
- You documented any public code that the end-user might use.
- You followed the [contributing file] 

### Type

- [ ] Internal change (Doesn't affect end-user).
- [ ] External change (Does affect end-user).
- [ ] Wiki (Changes towards the [Wiki]).
- [x] Other: Contributing.md file 

This refactors the CONTRIBUTING.md file with new, more detailed information.  
Additionally was the Code of Conduct part moved to its own dedicated file.

Please let me know about any improvements, additions, etc. that can be done to this files. I honestly struggle to find a proper way to make a contributing file.

[Pull requests]: https://github.com/PlaceholderAPI/PlaceholderAPI/pulls
[contributing file]: https://github.com/PlaceholderAPI/PlaceholderAPI/tree/master/.github/CONTRIBUTING.md
[Wiki]: https://github.com/PlaceholderAPI/PlaceholderAPI/wiki

PR #830, opened 2022-05-17
Slightly optimize replacer process

montlikadani:
### Type

- [ ] Other: __________ 

### Description

_According to CONTRIBUTING, changes should have been made to the `development` branch, but most of the PRs merged with the `master`, so I did committed here._

This PR is mainly a performance optimization that will have a low impact, so we can say that it is also a slightly faster implementation process. The following has changed, which is worth applying everywhere in the entire source code:

- The number of elements in `String#toCharArray` will be equal to `String#length`, so we can use the `Array#length` property because it takes a little longer to execute the `String#length` method in newer versions of Java.
- It is a good idea to declare some field types outside the loop if you want to use them more than once in a loop so that the compiler does not have to re-create a new field each time it is called. This applies to each type of primitives and wrappers.
- Unnecessary declarations, such as those removed in line 88, should not be declared if they are used only once.

If these changes don’t seem appropriate or meaningful, I can implement a benchmark if necessary. If appropriate with these changes, an optimal performance improvement could be achieved in the future (if there will be more similar improvements).

I committed this change because I recently received a performance report, which is obviously a problem with my resource, because the `PlaceholderAPI#setPlaceholders` method is constantly called on a synchronous thread, which degrades performance. With this change, we can make a small improvement regarding with replacers.


emilyy-dev:
> It is a good idea to declare some field types outside the loop if you want to use them more than once in a loop so that the compiler does not have to re-create a new field each time it is called. This applies to each type of primitives and wrappers.

Not to sound condescending but, this change is completely pointless. That is not the case, at all, the compiler doesn't create fields for those, it reserves the needed space in the LVT (for then the JVM to allocate it when it enters the function and for all of the stack to work nicely without crashing), but even if you do mean that, the LVT isn't changed in either case (the number of local variables needed is the same in both) and the generated bytecode is effectively the exact same line by line. There is no performance gain or loss, only readability matter.

src/main/java/me/clip/placeholderapi/replacer/CharsReplacer.java
```diff
@@ -104,7 +108,7 @@ public String apply(@NotNull final String text, @Nullable final OfflinePlayer pl
         continue;
       }
 
-      final PlaceholderExpansion placeholder = lookup.apply(lowercaseIdentifierString);
+      final PlaceholderExpansion placeholder = lookup.apply(identifierString.toLowerCase(Locale.ROOT));
       if (placeholder == null) {
         builder.append(closure.head).append(identifierString);
```
Starmism, line 114: What Emily said, this only hurts readability and doesn't actually improve anything performance-wise.

src/main/java/me/clip/placeholderapi/replacer/RegexReplacer.java
```diff
@@ -51,17 +51,15 @@ public String apply(@NotNull final String text, @Nullable final OfflinePlayer pl
...
-      final String parameters = matcher.group("parameters");
-
-      final PlaceholderExpansion expansion = lookup.apply(identifier);
-      if (expansion == null) {
+      if ((expansion = lookup.apply(matcher.group("identifier"))) == null) {
         continue;
       }
 
-      final String requested = expansion.onRequest(player, parameters);
+      requested = expansion.onRequest(player, matcher.group("parameters"));
```
Starmism, line 62: Same here

src/main/java/me/clip/placeholderapi/replacer/CharsReplacer.java
```diff
@@ -43,22 +43,27 @@ public CharsReplacer(@NotNull final Closure closure) {
   public String apply(@NotNull final String text, @Nullable final OfflinePlayer player,
       @NotNull final Function<String, @Nullable PlaceholderExpansion> lookup) {
     final char[] chars = text.toCharArray();
-    final StringBuilder builder = new StringBuilder(text.length());
+    final StringBuilder builder = new StringBuilder(chars.length);
```
Starmism, line 46: This is the only thing that could possibly be an improvement.

Starmism (changes requested):

src/main/java/me/clip/placeholderapi/replacer/CharsReplacer.java
montlikadani: What about if the above condition `invalid` is true, the iteration continues to the next element and the `String#toLowerCase` was unnecessarily called before? Useless calls such as this should be avoided.

src/main/java/me/clip/placeholderapi/replacer/RegexReplacer.java
Andre601: The regex replacer isn't even used, so this is completely pointless to "improve". If you want to improve it, remove it completely.

src/main/java/me/clip/placeholderapi/replacer/CharsReplacer.java
Andre601: Is the performance improvement really THAT significant?
If this is just a few ms, then changing this absolutely pointless and makes the code worse than it already is IMO.

Andre601:
> According to CONTRIBUTING, changes should have been made to the `development` branch, but most of the PRs merged with the `master`, so I did committed here.

You may not have noticed this, but there is an open PR called "Update Contributing file" (#407) and in that PR is the move away from `development` back to the `master` branch explained, including all changes to the contributing file.

The PR has been stuck in limbo, because nobody - primarily @darbyjack to which I replied - cares about my responses enough it seems... Or are to busy... idk.

src/main/java/me/clip/placeholderapi/replacer/CharsReplacer.java
Starmism: Yeah we're talking in literally like dozens of CPU cycles here, this changes absolutely nothing. Having the placeholders running on a separate thread would be a significantly bigger improvement than the micro optimization of this. Weighing the purported performance benefit vs the way it makes the code worse and harder to read, it is easy to make the call to leave it as is.

montlikadani: If the `invalid` condition is true in all cases and this method is called more than once, I could say yes.

montlikadani:
I would just remove the `development` branch in place after merging #773 so as not to disturb others.

Andre601:
> I would just remove the `development` branch in place after merging #773 so as not to disturb others.

Well... That's the plan obviously. But I'm still waiting for final replies on the PR, as I don't want to just blindly merge stuff, even tho I could.

PiggyPiglet:
The changes in this PR are superficial. We also do not wish to remove the regex replacer and its tests as of yet.

PR #859, opened 2022-07-18
Fix for a strange error that didn't load expansions without `/papi reload`

wxipwastaken:
### Type
- [ ] Other: __________ 

### Description
Fixes a strange error that didn't load expansions without `/papi reload`


Andre601:
What is the actual issue? Is it on startup?

wxipwastaken:
> What is the actual issue? Is it on startup?

Yes, the error was popping up at startup. [This was the error (before my "fix")](https://pastebin.com/n46Wxdsn)
Basically it wouldn't load any extensions until I type `/papi reload`.

PR #866, opened 2022-07-30
Fix duplicate expansion loading

### LINKED ISSUE #864: Broken expansion loading
### Confirmation

- [X] My issue isn't already found on the Issue tracker.
- [X] My issue is about **PlaceholderAPI** and not any expansion or external plugin
- [X] The issue isn't already fixed in a Spigot Release or Development Build.
- [X] The [Common Issues](https://github.com/PlaceholderAPI/PlaceholderAPI/wiki/Common-Issues) page doesn't mention this issue.

### Type

API Bug

### What happens?

Right now is PlaceholderAPI's expansion loading broken in a sense that it may load the same expansion twice, if there are 2 jar files of it.

This is partially to blame on how the expansion is downloaded. PlaceholderAPI will download the expansion into a jar named `expansion-<name>.jar`, replacing it, if already present. If there is the same expansion now, but with a different jar name will it not be replaced.

### Expected Behaviour

There should be a way to prevent the same expansion from loading twice, which could be achieved by checking if an expansion with that name already exists.

The jar replacement would be a more difficult story here.

### How to Reproduce

1. Manually download an expansion, or rename the jar already present
2. Download the same expansion through the `/papi ecloud` command.
3. Reload PlaceholderAPI.

### `/papi dump` Output

No real need for it.

### Console Log

_No response_

### Errors

_No response_

### Additional Info

_No response_

Andre601:
### Type

- [ ] Other: __________ 

### Description

Closes #864 

Fixes an issue where the same expansion would be loaded twice, if the jar file used has a slightly different name.

Now PlaceholderAPI checks if the used identifier is already registered and in such a case prints a warning and doesn't load the expansion.

### DIFF
--- a/src/main/java/me/clip/placeholderapi/expansion/manager/CloudExpansionManager.java
+++ b/src/main/java/me/clip/placeholderapi/expansion/manager/CloudExpansionManager.java
@@ -53,6 +53,7 @@ import java.util.stream.Collectors;
 import me.clip.placeholderapi.PlaceholderAPIPlugin;
 import me.clip.placeholderapi.expansion.PlaceholderExpansion;
 import me.clip.placeholderapi.expansion.cloud.CloudExpansion;
+import me.clip.placeholderapi.util.Msg;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Unmodifiable;
 
@@ -268,8 +269,7 @@ public final class CloudExpansionManager {
       await.remove(toIndexName(expansion));
 
       if (exception != null) {
-        plugin.getLogger().log(Level.SEVERE,
-            "failed to download " + expansion.getName() + ":" + version.getVersion(), exception);
+        Msg.severe("Failed to download %s:%s", exception, expansion.getName(), expansion.getVersion());
       }
     }, ASYNC_EXECUTOR);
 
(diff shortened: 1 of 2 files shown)


BlitzOffline (approved):
The only thing I'd really change is the message.

From `Failed to load expansion %s. Identifier is already registered.`
To `Failed to load expansion %s because another expansion with the same identifier is already registered.`

Since the identifier isn't registered, the expansion with that identifier is.

Andre601:
Feels s bit word...
Perhaps `Failed to load expansion '%s'. Identifier '%s' is already in use.`

Starmism (approved):

BlitzOffline (approved):

darbyjack (approved):

Frcsty:
Request to revert this merge as the actual identifier for the expansion should never be changed. Which in turn causes stuff to break when reloading.

Andre601:
> Request to revert this merge as the actual identifier for the expansion should never be changed. Which in turn causes stuff to break when reloading.

?

iGabyTM:
> Request to revert this merge as the actual identifier for the expansion should never be changed. Which in turn causes stuff to break when reloading.

Do you have proof that reload doesn't work or is just a guess? Iirc the expansions map is cleared on reload.

Andre601:
The issue in your case would more be doing stuff that nobody encourages.

Reloading the server (or in your case individual plugins) is something no plugin dev would recommend to you. And the issues you describe are one reason why.

PlaceholderAPI doesn't monitor plugins that register expansions to then unload them when the plugin itself shuts down. This at most happens with expansions that defined a required plugin (See [here](https://github.com/PlaceholderAPI/PlaceholderAPI/blob/81ef464dad3f58f758299cbd27e2af921e0bb570/src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java#L466-L482)).

The easiest way to fix this is to simply have the plugin unregister its expansion on disable to then register it again.

PR #923, opened 2023-02-26
Since .persist() is to indicate that the expansion is not provided th…

vk2gpz:
…rough eCloud but through a plugin, if this is true, there is no need to check against eCloud entries.

### Type

- [x ] Internal change (Doesn't affect end-user).
- [ ] Other: __________ 

### Description

Since .persist() is to indicate that the expansion is not provided through eCloud but through a plugin, if this is true, there is no need to check against eCloud entries.

### DIFF
--- a/src/main/java/me/clip/placeholderapi/expansion/manager/CloudExpansionManager.java
+++ b/src/main/java/me/clip/placeholderapi/expansion/manager/CloudExpansionManager.java
@@ -220,7 +220,7 @@ public final class CloudExpansionManager {
                             plugin.getLocalExpansionManager().findExpansionByName(name);
                         if (localOpt.isPresent()) {
                           PlaceholderExpansion local = localOpt.get();
-                          if (local.isRegistered()) {
+                          if (local.isRegistered() && !local.persist()) {
                             expansion.setHasExpansion(true);
                             expansion.setShouldUpdate(
                                 !local.getVersion().equalsIgnoreCase(expansion.getLatestVersion()));


Andre601:
I doubt it's only for those included in a plugin. Sure, it's the main use, but I wouldn't be surprised if some expansions exist that use this method too for whatever reason...
Just my opinion tho.

vk2gpz:
Yes, one scenario I can think of is that someone started an expansion as the plugin-provided, but later move to eCloud, and the eCloud version is newer.

The best solution would be to implement a proper version number check rather than using equalsIgnoreCase.  Then we don't need to check local.persist() and rely on proper version check.

However, in order for the version number check to work properly, PlaceholderAPI team should set the version number convention so that PAPI can implement a proper version check.

vk2gpz:
Once a proper version number checking is implemented '!local.persist()' can be removed and replace 'equalIgnoreCase()' with the new version checking method.

Andre601:
> However, in order for the version number check to work properly, PlaceholderAPI team should set the version number convention so that PAPI can implement a proper version check.

I don't think it's needed too much. Most Expansions I know of use a proper SemVer format (or SemVer-like).
And for the checks could a simple lib like [FlexVer](https://github.com/unascribed/FlexVer) be used to allow a more flexible version check to be done here.

Alternatively could we put the version check in the hands of the expansion devs. Like have a boolean method that they can override, where they could set up whatever checks they want to do to tell PAPI that a new version is available.

vk2gpz:
If PAPI is going to dump this on devs, then PAPI needs to include checkVersion(versionStr1, versionStr2) method for the expansion class to implement.

vk2gpz:
> > However, in order for the version number check to work properly, PlaceholderAPI team should set the version number convention so that PAPI can implement a proper version check.
> 
> I don't think it's needed too much. Most Expansions I know of use a proper SemVer format (or SemVer-like). And for the checks could a simple lib like [FlexVer](https://github.com/unascribed/FlexVer) be used to allow a more flexible version check to be done here.
> 
> Alternatively could we put the version check in the hands of the expansion devs. Like have a boolean method that they can override, where they could set up whatever checks they want to do to tell PAPI that a new version is available.

"most"... you should not make such an assumption.  when it comes to version numbering, one can define anything.
x.x.x, x.x.x-RCx, x.x.x.build#, x.x-beta, x.x.x-alpha....

vk2gpz:
or should separate the process of checking update is needed and do not include in the process of generating the elcoud installed list.

Andre601:
> "most"... you should not make such an assumption. when it comes to version numbering, one can define anything.
> x.x.x, x.x.x-RCx, x.x.x.build#, x.x-beta, x.x.x-alpha....
> If you guys think some lib like FlexVer is good enough, you should implement it instead of equalsIgnoreCase.

FlexVer would be a lib to cover the most common cases here like the ones you mentioned, as it was made for MC mods in mind, which have all kinds of... *interesting* formats for their versions like f.e. `1.2.3+1.19.3-b1`

We can't cover all cases. That's literally impossible and the biggest downside of having the version be a String instead of an integer, double or similar, which at the very least would force it to be a number and would therefore have a basic level of consistency.

That's why I do also suggest the option of allowing the expansion itself to tell PAPI if an update is newer or not. And it would be like you suggested in your other comment.

The only real other fix I can imagine for the local (part of plugin) vs cloud expansion issue is to have some way for the dev to tell PAPI that their expansion is part of a plugin and not the ecloud (i.e. a getter they can override that returns a boolean value).
Issue with that is, that it again depends on the dev to do the work, which most will not bother with.

Another idea could be to make the register method plugins use set this particular expansion as local, while using an internal method for the cloud ones. This could be the better aproach here probably, thos since that method is part of the PlaceholderExpansion class would it be exposed either way, so a dev could use it and trick PAPI into thinking it's a cloud expansion while it isn't...

There are many such issues we need to look into for the future.

vk2gpz:
> 
> The only real other fix I can imagine for the local (part of plugin) vs cloud expansion issue is to have some way for the dev to tell PAPI that their expansion is part of a plugin and not the ecloud (i.e. a getter they can override that returns a boolean value). Issue with that is, that it again depends on the dev to do the work, which most will not bother with.
>

According to the document, this is exactly "persist()"'s purpose, that's why I originally submitted this PR. but I guess I will let you guys decide what to do.

PR #929, opened 2023-03-10
Added AbstractPlaceholderExpansion for more readable expansion.

DavidTheExplorer:
### Type

- [ ] Other: __________ 

### Description

This PR adds a convenient base class for **PlaceholderExpansion** with constructor, so people no longer need to implement _getVersion(), getIdentifier(), getAuthor()_ as they're annoying as hell to read.

Example from my actual plugin:
image

### DIFF
--- a/src/main/java/me/clip/placeholderapi/expansion/AbstractPlaceholderExpansion.java
+++ b/src/main/java/me/clip/placeholderapi/expansion/AbstractPlaceholderExpansion.java
@@ -1,0 +1,30 @@
+package me.clip.placeholderapi.expansion;
+
+/**
+ * This is a convenient skeletal class for all plugins that want to supply placeholders.
+ */
+public abstract class AbstractPlaceholderExpansion extends PlaceholderExpansion
+{
+	private final String identifier, version, author;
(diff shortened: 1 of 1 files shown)


Andre601:
I myself can't really see the overall benefit of this...
Like how is a `super("identifier", "version", "author")` any better to read than the getters?

IMO do the getters actually help clarify what the value returned is: The identifer, the version or the author. With the constructor, it's more difficult to know what is what exactly.

DavidTheExplorer:
> I myself can't really see the overall benefit of this... Like how is a `super("identifier", "version", "author")` any better to read than the getters?
> 
> IMO do the getters actually help clarify what the value returned is: The identifer, the version or the author. With the constructor, it's more difficult to know what is what exactly.
> 
> Tho, obviously is this just my opinion.
Nothing is changed functionality-wise. It makes no sense to enforce 3 getters instead of a constructor for no reason, the constructor makes the expansion class way more concise and readable.
Also the constructor itself isn't really that hard to understand...

DavidTheExplorer:
?

PR #944, opened 2023-03-27
[Wiki] Improve PlaceholderExpansion Page

Andre601:
### Type

- [ ] Other: __________ 

### Description

Closes N/A 

Improves the PlaceholderExpansion page for the wiki by recommending the internal use of an expansion over external use.

### DIFF
--- a/PlaceholderExpansion.md
+++ b/PlaceholderExpansion.md
@@ -1,12 +1,8 @@
 [placeholderexpansion]: https://github.com/PlaceholderAPI/PlaceholderAPI/blob/master/src/main/java/me/clip/placeholderapi/expansion/PlaceholderExpansion.java
-
-[playerexpansion]: https://github.com/PlaceholderAPI/Player-Expansion
-[serverexpansion]: https://github.com/PlaceholderAPI/Server-Expansion
-[mathexpansion]: https://github.com/Andre601/Math-expansion
-
 [relational]: https://github.com/PlaceholderAPI/PlaceholderAPI/blob/master/src/main/java/me/clip/placeholderapi/expansion/Relational.java
 
 ## Overview
+
 This page will cover how you can create your own [`PlaceholderExpansion`][placeholderexpansion] which you can either [[Upload to the eCloud|Expansion cloud]] or integrate into your own plugin.
 
 It's worth noting that PlaceholderAPI relies on expansions being installed. PlaceholderAPI only acts as the core replacing utility while the expansions allow other plugins to use any installed placeholder in their own messages.
@@ -16,21 +12,28 @@ You can download Expansions either directly from the eCloud yourself, or downloa
 
 - [Getting started](#getting-started)
   - [Common Parts](#common-parts)
-- [Without a Plugin](#without-a-plugin)
-- [With a Plugin (External Jar)](#with-a-plugin-external-jar)
-- [With a Plugin (Internal Class)](#with-a-plugin-internal-class)
-  - [Register the Expansion](#register-the-expansion)
+- [Making an internal Expansion](#making-an-internal-expansion)
+  - [Full Example](#full-example)
+  - [Register your Expansion](#register-your-expansion)
+- [Making an external Expansion](#making-an-external-expansion)
+  - [Full Example (Without Dependencies)](#full-example-without-dependencies)
+  - [Full Example (With Dependencies)](#full-example-with-dependencies)
 - [Relational Placeholders](#relational-placeholders)
-  - [Notes about Relational Placeholders](#notes-about-relational-placeholders)
+  - [Quick Notes](#quick-notes)
+  - [Adding Relational Placeholders](#adding-relational-placeholders)
+  - [Full Example](#full-example-1)
 
 ## Getting started
+
 For starters, you need to decide what type of [`PlaceholderExpansion`][placeholderexpansion] you want to create. There are various ways to create an expansion. This page will cover the most common ones.
 
 ### Common Parts
+
 All shown examples will share the same common parts that belong to the [`PlaceholderExpansion`][placeholderexpansion] class.  
 In order to not repeat the same basic info for each method throughout this page, and to greatly reduce its overall length, we will cover the most basic/necessary ones here.
 
 #### Basic PlaceholderExpansion Structure
+
 ```java
 package at.helpch.placeholderapi.example.expansions;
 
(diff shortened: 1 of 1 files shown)


PlaceholderExpansion.md
```diff
@@ -71,265 +74,298 @@ Let's quickly break down the different methods you have to implement.
...
+
+## Making an external Expansion
+
+External expansions are separate Jar files containing the PlaceholderExpansion extending class.  
+They are recommended for the following scenarios:
+
+- Your Expansion doesn't depend on any Plugin.
+- Your Expansion depends on a Plugin AND you can't directly implement the Class into it.
+
+For Expansions made for a plugin is it recommended to try and use [internal Expansions](#making-an-internal-expansion). If you can't implement it yourself, i.e. because you're not the dev of it and it isn't open source, should you consider alternative solutions (i.e. asking the dev to implement it) before using this method.
```
WiIIiam278, line 214: May I suggest adding something even more obvious than this to below the header line?

```
> ⚠️ If you'd like to provide PlaceholderAPI expansions for your own plugin, this is not recommended. Instead, [register an internal expansion provider](#making-an-internal-expansion).
```

PlaceholderExpansion.md
```diff
@@ -71,265 +74,298 @@ Let's quickly break down the different methods you have to implement.
...
-When creating such an expansion is it recommended to use [`onRequest(OfflinePlayer, String)`](#onrequestofflineplayer-string).
+It is important to note that PlaceholderExpansions loaded manually through the `register()` method (Therefore not being loaded by PlaceholderAPI itself) **require** to have the `persist()` method return `true` to avoid the expansion being unloaded during a `/papi reload`.
 
-#### Full Example
-Please see the [Common parts](#common-parts) section for info on the other methods.
+Depending on what info you try to access from your plugin (i.e. through the main class or a dedicated API) can you use Dependency Injection to obtain an instance of whatever you need.  
+Keep in mind that as mentioned earlier, the Expansion class won't be reloaded when `persist()` is set to `true`, so you have to refresh any info that may get outdated during a reload yourself.
 
+Here is an example of a possible Dependency Injection:  
 ```java
```
WiIIiam278, line 101: Some of these examples can get quite long -- perhaps surround with GitHub's collapsible detail elements to make reading this a little easier?

```
<details>
<summary>Example registration through dependency inejction</summary>

\```java
#foo
\```

</details>
```

WiIIiam278 (approved):
On the double, this is definitely an improvement, though I definitely still think it could be made even more clear _not_ to create external expansions if you're the plugin developer, as they really are intended more for developers wishing to provide additional placeholders for existing other projects.

I also think this would be better split into multiple pages given its length;
- the introduction and common elements
- registering an internal expansion provider
- creating an external PAPI expansion (and also some docs on how to add it to the eCloud, as currently that's difficult as restrictions apply to account creation & approval, and it'd be good to document those as it's a bit of a pitfall atm)

PlaceholderExpansion.md
Andre601: I feel like this would be too biased here.

It already states the main cases this type of expansion can be used for, so I think that's fine enough.

WiIIiam278: It does. Just need a space between the HTML tag and markdown

PlaceholderExpansion.md
WiIIiam278: It does. Just need a space between the HTML tag and markdown

PlaceholderExpansion.md
Starmism: Yeah it should really emphasize that 99.99% of the time, you want to make an internal expansion. Also would be good to add the benefit that you don't need to match versions between the expansion and the plugin, and that setup is easier since it doesn't require downloading from the ecloud.

PlaceholderExpansion.md
Andre601: Moved code examples and the explanation of the methods into details. I kinda dislike that details are barely noticable, but I think it should be fine.

PlaceholderExpansion.md
```diff
@@ -71,265 +76,314 @@ Let's quickly break down the different methods you have to implement.
...
 
-----
-## With a Plugin (External Jar)
-If your expansion relies on a plugin to provide its placeholder values, you will need to override a few more methods to make sure everything will work correctly.
+</p>
+</details>
 
-Your expansion will need to override the `getRequiredPlugin()` method to return the name of the plugin your expansion depends on.  
-PlaceholderAPI automatically checks if this method will either return null, or if the name defined results in a non-null plugin.
+### Register your expansion
```
PiggyPiglet, line 186: your -> Your
expansion -> Expansion

PiggyPiglet (approved):
good to merge, my comments are just on capitalisation which doesn't really matter, idc if the things I highlighted get changed or not

WiIIiam278:
(I still think you should more explicitly warn against making external expansions if you’re the plugin dev as per previous review!)

Andre601:
> (I still think you should more explicitly warn against making external expansions if you’re the plugin dev as per previous review!)

I'm personally against that. While internal expansions may have benefits is making an external one not inherently bad. So "warning" them about that would be the wrong aproach.
The current setup should imo be the best aproach here, as it explains internal ones first with their benefits before later tackling external ones.

WiIIiam278:
> > (I still think you should more explicitly warn against making external expansions if you’re the plugin dev as per previous review!)
> 
> I'm personally against that. While internal expansions may have benefits is making an external one not inherently bad. So "warning" them about that would be the wrong aproach. 

I don’t have metrics on hand but I’d wager 85%+ of the developers reading this bit of documentation are looking to add placeholders for their plugin. I just don’t see a compelling argument for suggesting (accidentally or otherwise) that developers undertake the additional effort to make and distribute a jar on a site that has account creation closed and requires pinging support on a discord whenever they wish to distribute an update to verify it, when it would be far easier for them to build an expansion into the jar file.

If a warning isn’t what you’d like to add, maybe some examples? Like “Who is an external expansion good for?” —> Bespoke / independent extra placeholders, add-ons for existing plugins made by other authors VS. “Who is an internal expansion good for?” —> If you want to add placeholder support for your own plugin.

> The current setup should imo be the best aproach here, as it explains internal ones first with their benefits before later tackling external ones.

Also want to add: The first paragraph presents making an external expansion as the first option:
> This page will cover how you can create your own [`PlaceholderExpansion`][placeholderexpansion] which you can either [[Upload to the eCloud|Expansion cloud]] or integrate into your own plugin

At the very least I think care should be taken to present an external expansion as the second option in the wording and language

Andre601:
> If a warning isn’t what you’d like to add, maybe some examples? Like “Who is an external expansion good for?” —> Bespoke / independent extra placeholders, add-ons for existing plugins made by other authors VS. “Who is an internal expansion good for?” —> If you want to add placeholder support for your own plugin.


[Lines 217-223](https://github.com/PlaceholderAPI/PlaceholderAPI/pull/944/files#diff-2a83502f65b098dec3947517c565b08dffce07926cf99240f056e1a6e12ed6c1R217-R223):
> External expansions are separate Jar files containing the PlaceholderExpansion extending class.  
> They are recommended for the following scenarios:
> 
> - Your Expansion doesn't depend on any Plugin.
> - Your Expansion depends on a Plugin AND you can't directly implement the Class into it.
> 
> If you'd like to implement PlaceholderAPI expansions for your own plugin consider using the [internal expansion setup](#making-an-internal-expansion) instead.

WiIIiam278:
> [Lines 217-223](https://github.com/PlaceholderAPI/PlaceholderAPI/pull/944/files#diff-2a83502f65b098dec3947517c565b08dffce07926cf99240f056e1a6e12ed6c1R217-R223):

yeah okay, I do see this now :) thanks for highlighting it to me. I did give it a read through; my eyes must’ve glanced over this.

> I honestly won't have this discussion here, as this is just pointless nitpicking that would barely give any benefits whatsoever

No offence intended!! All in good heart from my end :) — your work here is of course appreciated. 

if you’d like you can read through the discussion we had in the discord a few months ago, where we largely reached the same conclusion.

> I will assume and hope that the common reader will be able to understand the overall benefits and drawbacks of each type and won't handhold them throughout the entire thing

I think that’s a reasonable assumption in general; I’m just wary as my experienced using the ecloud was rather messy to say the least and I was rather ruffled when I was ultimately told by staff that I had made a weird choice by not just building my expansion into my plugin.

Andre601:
Perhaps the ecloud itself should be updated and have like a large banner informing the user that it's recommended to try and make internal expansions... Or to mention the most common types of expansions this ecloud is meant for.

I just feel like the wiki itself shouldn't be the only source here and we should also stay away from being so opinionated here. Each type has its benefits and drawbacks and we shouldn't tell users "This one is much more awesome!" tbh...

Andre601:
I close this in favour of #1032 for a better wiki experience.

PR #953, opened 2023-04-05
feat: distinguish expansions (external or internal) (fixes #945)

### LINKED ISSUE #945: Consider alternative solution for #864
### Confirmation

- [X] My issue isn't already found on the Issue tracker.
- [X] My issue is about **PlaceholderAPI** and not any expansion or external plugin
- [X] The issue isn't already fixed in a Spigot Release or Development Build.
- [X] The [Common Issues](https://github.com/PlaceholderAPI/PlaceholderAPI/wiki/Common-Issues) page doesn't mention this issue.

### Type

Plugin Bug

### What happens?

Reloading a plugin that provides an expansion through [ServerUtils](https://github.com/FrankHeijden/ServerUtils) fails due to PAPI refusing to load an expansion when one with the same identifier is already registered. Prior to 2.11.3 release, specifically #866, this worked. 

### Expected Behaviour

Ideally, both this behavior is retained, and #864 is resolved

### How to Reproduce

1. Load a server with ServerUtils, PlaceholderAPI (2.11.3), and a plugin that provides a PlaceholderExpansion
2. `/su rp <that plugin>`

### `/papi dump` Output
N/A (Paper 1.19.2, PlaceholderAPI 2.11.3)

### Console Log
N/A

### Errors
`[21:54:57 WARN]: [PlaceholderAPI] Failed to load expansion <that plugin's expansion's identifier>. Identifier is already in use.`

### Additional Info

After 2.11.3 release, specifically #866, we started experiencing issues with reloading (through ServerUtils) plugins that provide an expansion. It's understood that reloading plugins at runtime is generally unsafe and not supported, but before this PR was merged, it was possible to reload plugins without issues from PAPI. As said above, ideally both this behavior is retained and #864 is resolved.

I'm not sure I understand the true issue in #864. My understanding is: PlaceholderAPI may attempt to register an expansion twice, but because expansions are backed by a map with their identifier as the key, it'll keep and reference the latest one. For this issue, this is ideal behavior. 

There are a few possible solutions I have in mind:
- Warn the user that an expansion is being registered and overriding the already loaded one
- All plugins unregister their own expansions (while already possible, I believe it is not suggested in PAPI'S docs)

iGabyTM:
external: located in the `expansions` folder
internal: registered by a plugin

### Type

- [ ] Other: __________ 

### Description

Closes N/A 

### DIFF
--- a/src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
+++ b/src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
@@ -294,21 +294,24 @@ public final class LocalExpansionManager implements Listener {
       Bukkit.getPluginManager().registerEvents(((Listener) expansion), plugin);
     }
     
-    Msg.info("Successfully registered expansion: %s [%s]", expansion.getIdentifier(),
-        expansion.getVersion());
+    Msg.info(
+            "Successfully registered %s expansion: %s [%s]",
+            expansion.getExpansionType().name().toLowerCase(),
+            expansion.getIdentifier(),
+            expansion.getVersion()
+    );
 
     if (expansion instanceof Taskable) {
       ((Taskable) expansion).start();
     }
 
-    if (plugin.getPlaceholderAPIConfig().isCloudEnabled()) {
-      final Optional<CloudExpansion> cloudExpansionOptional =
-          plugin.getCloudExpansionManager().findCloudExpansionByName(identifier);
+    // Check eCloud for updates only if the expansion is external
+    if (plugin.getPlaceholderAPIConfig().isCloudEnabled() && expansion.getExpansionType() == PlaceholderExpansion.Type.EXTERNAL) {
+      final Optional<CloudExpansion> cloudExpansionOptional = plugin.getCloudExpansionManager().findCloudExpansionByName(identifier);
       if (cloudExpansionOptional.isPresent()) {
         CloudExpansion cloudExpansion = cloudExpansionOptional.get();
         cloudExpansion.setHasExpansion(true);
-        cloudExpansion.setShouldUpdate(
-            !cloudExpansion.getLatestVersion().equals(expansion.getVersion()));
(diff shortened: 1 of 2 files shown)


Starmism (approved):

Andre601:
Would it be possible to expose the boolean in some way for internal use?

It would allow us to check if the expansion is internal or external for stuff like checking for Updates on the ecloud (Avoiding false positives)

src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
```diff
@@ -308,6 +315,19 @@ public boolean register(@NotNull final PlaceholderExpansion expansion) {
...
   }
 
+  /**
+   * Overload for {@link #register(PlaceholderExpansion, boolean)} to provide backwards compatibility for expansions / plugins
+   * that call this method directly. It is the equivalent of {@code register(expansion, false)}
+   * @param expansion the expansion to register
+   * @return if the expansion was registered
+   * @deprecated use {@link #register(PlaceholderExpansion, boolean)} directly
+   */
+  @Deprecated
```
Andre601, line 325: Do we actually need this? Since it's marked as Internal API usage is it imo not really necessary to add docs and also deprecate what is supposed to be only used by us.

Andre601:
Line 304 should have a `&& isExternalExpansion` added to avoid checking for updates with a local expansion.

src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
iGabyTM: I mean, there's nothing wrong with documenting and deprecating internal use methods.

iGabyTM:
> Line 304 in LocalExpansionManager should have a `&& isExternalExpansion` added to avoid checking for updates with a local expansion.

Done

Andre601 (approved):

BlitzOffline (approved):

PiggyPiglet (approved):

PR #958, opened 2023-04-19
Make REPLACER_PERCENT and REPLACER_BRACKET available for hijacking re…

armagidon-exception:
### Type
- [ ] Other: ______

### Description
PlaceholderAPI does not allow to hijack requests for replacement. This PR adds new event PlaceholderRequestEvent that is fired whenever a user requests for placeholders. Developer then can change what result will be produced for this request.

### DIFF
--- a/src/main/java/me/clip/placeholderapi/PlaceholderAPI.java
+++ b/src/main/java/me/clip/placeholderapi/PlaceholderAPI.java
@@ -43,8 +43,8 @@ import org.jetbrains.annotations.NotNull;
 
 public final class PlaceholderAPI {
 
-  private static final Replacer REPLACER_PERCENT = new CharsReplacer(Closure.PERCENT);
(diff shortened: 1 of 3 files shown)


Andre601:
What would be the actual use case here?
Like I myself can't see a real justification for allowing a plugin to just override any text PAPI parses.

If a plugin parses `Hello %player_name%` but receives a completely different message because another plugin altered is that imo bad, since you expect some specific outputs.

armagidon-exception:
> What would be the actual use case here? Like I myself can't see a real justification for allowing a plugin to just override any text PAPI parses.
> 
> If a plugin parses `Hello %player_name%` but receives a completely different message because another plugin altered is that imo bad, since you expect some specific outputs.

The use-case is that if you want to integrate your own system but don't want to lose compatibility with other plugins who use papi and not your system, with this event, you can parse placeholders differently, however you want to.

PR #978, opened 2023-06-23
Add environment variable to override cloud_allowunverified_expansions

JulianVennen:
### Type
- [ ] Other: __________ 

### Description
This pull request adds the environment variable PAPI_ALLOW_UNVERIFIED_EXPANSIONS.
It allows overriding the option set in the config.yml.

~~Accepted values are (case-insensitive) `true`, `yes` and `1` to allow expansions or `false`, `no` and `0` to forbid them.
If the environment variable is not set or the value is not valid, the value specified in the config.yml will be used.~~

If the value is set to `true` unverified expansions will be allowed (case-insensitive), if it is set to any other value, unverified expansions will be forbidden.

### DIFF
--- a/src/main/java/me/clip/placeholderapi/commands/impl/cloud/CommandECloudDownload.java
+++ b/src/main/java/me/clip/placeholderapi/commands/impl/cloud/CommandECloudDownload.java
@@ -51,16 +51,7 @@ public final class CommandECloudDownload extends PlaceholderCommand {
   private boolean areUnverifiedExpansionsAllowed(@NotNull final PlaceholderAPIPlugin plugin) {
     String env = System.getenv("PAPI_ALLOW_UNVERIFIED_EXPANSIONS");
     if (env != null) {
-      switch (env.toLowerCase()) {
-        case "true":
-        case "yes":
-        case "1":
-          return true;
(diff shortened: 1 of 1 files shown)


Andre601:
Nice idea, tho I personally feel like supporting all these specific options for what is basically true/false is not needed.
Just allow true false and then use `env.equalsIgnoreCase("true")` to check.

JulianVennen:
> Nice idea, tho I personally feel like supporting all these specific options for what is basically true/false is not needed. Just allow true false and then use `env.equalsIgnoreCase("true")` to check.

That would also be ok with me, I could make the change to that if requested

iGabyTM:
There is also this shortcut https://docs.oracle.com/javase/8/docs/api/java/lang/Boolean.html#getBoolean-java.lang.String-
And I agree with Andre, true and false are enough.

Andre601:
> There is also this shortcut https://docs.oracle.com/javase/8/docs/api/java/lang/Boolean.html#getBoolean-java.lang.String-
> And I agree with Andre, true and false are enough.

That method is just `return value.equalsIgnoreCase("true");` xD

iGabyTM (approved):

PR #1000, opened 2023-08-28
Added missing NotNull annotations to setBracketPlaceholders methods

mfnalex:
### Type

- [X] Other: Added missing annotations, so it kinda does, but also does not really affect API users

### Description

This adds missing NotNull annotations to three setBracketPlaceholders methods. As they all internally just call methods which already are annotated with NotNull for both the return value and the parameters, it's safe to add the annotations to these methods too.

(In fact I was kinda surprised when those annotations were missing, so I looked at the source to confirm the annotations are just missing instead of being absent on purpose)


oliver194:
pull request 1000 hmm

iGabyTM (approved):

PR #1014, opened 2023-10-28
Warn and prevent old vulnerable expansions

TypicalModMaker:
### Type

- [ ] Other: __________ 

### Description

Warn and prevent users from using old expansions that pose a security vulnerability if threat actor has permission to trigger placeholders
In my opinion, this could help some networks prevent security issues
Any suggestions are highly appreciated since I know my code doesn't look good.


Andre601:
The idea is nice, but having the values hardcoded into the plugin is a bad idea.
Because every time there would be a new security issue with an expansion version would we need to update the code and publish a new version for people to download. This would just be bloat.

A better aproach would be to have an option in the ecloud's expansion version output and have PAPI check that.

TypicalModMaker:
> The idea is nice, but having the values hardcoded into the plugin is a bad idea. Because every time there would be a new security issue with an expansion version would we need to update the code and publish a new version for people to download. This would just be bloat.
> 
> A better aproach would be to have an option in the ecloud's expansion version output and have PAPI check that.

Well my intial idea was to ask pigplet or someone to make an endpoint at ecloud api but i doubt they will listen to me when i made all of those exploits ¯\_(ツ)_/¯

Andre601:
Not merging this as we now have applied some patches on the ecloud and in the plugin to hopefully avoid these security issues in the future.

PR #1022, opened 2023-11-26
[Wiki] Remove incorrect "optional" word

AntonMenov:
### Type

- [ ] Other: __________ 

### Description

This commit removes the incorrect word "optional" in _depend_ comment. Before this commit, both _softdepend_ and _depend_ had the same "optional" comment of.

Closes N/A 

### DIFF
--- a/Hook-into-PlaceholderAPI.md
+++ b/Hook-into-PlaceholderAPI.md
@@ -94,7 +94,7 @@ Next step is to go to your plugin.yml or paper-plugin.yml and add PlaceholderAPI
   author: author
   main: your.main.path.Here
   
-  # This sets PlaceholderAPI as an optional dependency for your plugin.
+  # This sets PlaceholderAPI as a dependency for your plugin.
   depend: [PlaceholderAPI]
   ```
 - `paper-plugin.yml`


Hook-into-PlaceholderAPI.md
```diff
@@ -94,7 +94,7 @@ Next step is to go to your plugin.yml or paper-plugin.yml and add PlaceholderAPI
   author: author
   main: your.main.path.Here
   
-  # This sets PlaceholderAPI as an optional dependency for your plugin.
+  # This sets PlaceholderAPI as a dependency for your plugin.
```
Andre601, line 97: What about adding that it makes it required?
Would that be too much?

AntonMenov: It's hard to say which is better. In my opinion, "mandatory" sounds better than "required", but perhaps it's still too much.

Andre601: Yeah. The current change is fine enough. We can consider a change for the future perhaps...

Andre601 (approved):

Andre601:
Thanks for the correction.

PR #980, opened 2023-06-26
Folia support

Anon8281:
### Type

- [x] Internal change (Doesn't affect end-user).
- [ ] External change (Does affect end-user).
- [ ] Wiki (Changes towards the [Wiki]).
- [ ] Other: __________ 

### Description
Added Folia support using https://github.com/Anon8281/UniversalScheduler

PR #1026, opened 2023-12-16
Supports both bukkit and folia

handy-git:
Supports both bukkit and folia

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -23,7 +23,6 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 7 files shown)


Andre601:
Are there any benefits compated to #980 which also tries to add Folia support?

handy-git:
This PR is based on https://github.com/handy-git/FoliaLib

Advantages: 
1. Added more support for bstats
2. It is more convenient to use static methods 

You can compare the two underlying projects to determine which one needs to be merged

PR #1032, opened 2024-01-15
[Proposal] Move Wiki to GitHub Pages using MkDocs

Andre601:
### Type

- [ ] Other: __________ 

### Description

This PR acts as a proposal to move the wiki from its current location in the wiki tab of this repository, to a dedicated page using GitHub Pages and the static site generator MkDocs.

Here are some reasons for why a move is recommended:

- Higher control over the themeing such as colors used.
- Better page organization, allowing to put pages into sub-categories (See placeholder list page for an example).
- (probably) better SEO by allowing more ways to optimize and change page meta.
- More extensive Markdown with the option to extend it even further (i.e. more admonition boxes, better looking details, etc.). For example does one extension allow direct linking of repos, issues, etc. using shorthand formats such as `?<id>` (for discussions on this repo) or `<user>/<repo>`

There are a few downsides tho:

- Markdown is more strict. For example do lists have some very strict rules, like requiring an empty line before the start of one to have it render properly and that indendet entries (including text on new lines) need 4 spaces.
- Pages can be added just fine, and can be accessed through URL, but won't be displayed in any navigation unless you include them in the `nav` option.
- Certain actions like per-page embeds (When linked on Discord) require overriding of the theme used to allow customization.

A live preview of these changes is available here: https://docs.andre601.ch/PlaceholderAPI

Apreciate any feedback on it.


Andre601:
I want to also mention that there would be certain steps required to be made in this repository, should this PR and the move be accepted:

- In the `Pages` setting would you need to set the source to `GitHub Actions`
- In the `Environments` setting would you need to update the github-pages environment (or make one if not present) to have the wiki branch set. This one is important as it will otherwise prevent the action from running.

Funny-cube (approved):

Issue #1070, opened 2024-07-07
Placeholder expansion registration initializing...


3328429240:
### Confirmation

- [X] My issue isn't already found on the Issue tracker.
- [X] My issue is about **PlaceholderAPI** and not any expansion or external plugin
- [X] The issue isn't already fixed in a Spigot Release or Development Build.
- [x] The [Common Issues](https://github.com/PlaceholderAPI/PlaceholderAPI/wiki/Common-Issues) page doesn't mention this issue.

### Type

Plugin Bug

### What happens?

After the server starts, the variables can be displayed and output correctly, but a few seconds after the server starts, it displays' Placeholder expansion registration initialization... ' After the words were written, a command similar to 'papi reload' was executed, which reloaded all variables, preventing the previously registered variables from being used properly.
After this information, the 'guide' variable that previously displayed 'PlaceholderAPI] Successfully registered internal expansion: guide [1.0.0]' in the log does not appear in the 'papi list' list, and it does not parse properly when used. Therefore, I suggest adding a switch to optionally perform the reload like operation after server startup. Thank you—— Google Translate

### Expected Behaviour

At the prompt 'Placeholder expansion registration initialization...' The previously registered variables can still be used afterwards

### How to Reproduce

Start server

### `/papi dump` Output

none

Andre601:
This is not an issue with PlaceholderAPI but with the Guide expansion. Let me explain.

Based on the info you gave is Guide an internal expansion, which means 2 things:
1. It's not a separate jar file you add to your expansions folder, but a class inside an existing plugin.
2. Said plugin is registering the expansion, not PlaceholderAPI.

And there is the issue. The plugin that contains the Guide expansion is registering it, not PlaceholderAPI. This comes with the downside, that during a reload of PlaceholderAPI, it will unregister all expansions, but only register those inside the expansions folder again, as those are the only ones it can know about (How should it know about other expansions in a plugin? It does not and cannot know what plugin registers the expansion).

That's why internal expansions **need** to override the `persist()` method and have it return `true`, as this tells PlaceholderAPI to not unload it during a reload, avoiding this issue.

So the issue is the fact that the Guide expansion is not overriding said persist method, making PlaceholderAPI unregister it during a reload. And since it is not an external expansion in the expansions folder can PlaceholderAPI not "know" about it and therefore not register it...

[...]

PR #1071, opened 2024-07-07
Use ExpansionType check for persistence

Andre601:
### Type

- [s] External change (Does affect end-user).
- [ ] Other: __________ 

### Description

Closes N/A 

PlaceholderAPI is still solely relying on the `persist()` boolean method to check if an expansion should be unregistered or not.
And as prooven by #1070 is this a faulty aproach, as a dev missing the override or not setting it to return true can cause an internal expansion to unregister and not register again due to it being internal.

Since we now have an Enum to set an expansion's type to be either Internal (Default) or external, is there no need for this extra method to be needed, hence why this PR deprecates it and updates the expansion manager to check not only for `persist()` being true, but also for when it is marked internal.

This also adds a warning for the case where an expansion is marked external, but has `persist()` return true, which it should never be.

### DIFF
--- a/src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
+++ b/src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
@@ -46,6 +46,7 @@ import me.clip.placeholderapi.expansion.Cacheable;
 import me.clip.placeholderapi.expansion.Cleanable;
 import me.clip.placeholderapi.expansion.Configurable;
 import me.clip.placeholderapi.expansion.PlaceholderExpansion;
+import me.clip.placeholderapi.expansion.PlaceholderExpansion.Type;
 import me.clip.placeholderapi.expansion.Taskable;
 import me.clip.placeholderapi.expansion.VersionSpecific;
 import me.clip.placeholderapi.expansion.cloud.CloudExpansion;
@@ -264,6 +265,9 @@ public final class LocalExpansionManager implements Listener {
 
     if (expansion instanceof VersionSpecific) {
       VersionSpecific nms = (VersionSpecific) expansion;
+      Msg.warn("Nag Author(s) %s of expansion %s about their usage of the deprecated " 
+          + "VersionSpecific interface!", expansion.getAuthor(), expansion.getIdentifier());
+      Msg.warn("They should switch to a new method of determining the Server version.");
       if (!nms.isCompatibleWith(PlaceholderAPIPlugin.getServerVersion())) {
         Msg.warn("Your server version is incompatible with expansion %s %s",
             expansion.getIdentifier(), expansion.getVersion());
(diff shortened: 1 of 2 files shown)


src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
```diff
@@ -323,6 +327,19 @@ public boolean unregister(@NotNull final PlaceholderExpansion expansion) {
     if (expansions.remove(expansion.getIdentifier().toLowerCase(Locale.ROOT)) == null) {
       return false;
     }
+    
+    // Don't unregister expansions that are marked internal or set to be persistent.
+    if (expansion.getExpansionType() == Type.INTERNAL || expansion.persist()) {
+      // Print warning if an external expansion is set to be persistent.
+      if (expansion.getExpansionType() == Type.EXTERNAL && expansion.persist()) {
+        Msg.warn("Nag author(s) %s about their expansion %s being marked as \"external\" "
```
BlitzOffline, line 335: The author didn't mark the expansion type and probably don't know what that means. The message should probably be changed to something more like "Nag author(s) %s about their external expansion having persist() set to true!"

src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
```diff
@@ -323,6 +327,19 @@ public boolean unregister(@NotNull final PlaceholderExpansion expansion) {
     if (expansions.remove(expansion.getIdentifier().toLowerCase(Locale.ROOT)) == null) {
       return false;
     }
+    
+    // Don't unregister expansions that are marked internal or set to be persistent.
+    if (expansion.getExpansionType() == Type.INTERNAL || expansion.persist()) {
```
BlitzOffline, line 332: Shouldn't this check be made before the removal from the map on line 327?

src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
```diff
@@ -393,7 +410,15 @@ private void registerAll(@NotNull final CommandSender sender) {
 
   private void unregisterAll() {
     for (final PlaceholderExpansion expansion : Sets.newHashSet(expansions.values())) {
-      if (expansion.persist()) {
+      if (expansion.persist() || expansion.getExpansionType() == Type.INTERNAL) {
```
BlitzOffline, line 413: Can't we not skip external expansions here, and just skip them in the uregister(PlaceholderExpansion) method? This way we don't have duplicate code. This would mean, we would only skip here if expansion type is internal.

BlitzOffline (changes requested):

src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
Andre601: true

src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
iGabyTM: +1

src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
Andre601: I decided to skip unregister for external if persist is true, because maybe the author didn't understand stuff and decided to override the expansion type when they shouldn't? Causing this particular situation of an external expansion being set to be persistent...

Perhaps we could make it different by still unregister external expansions, no matter the persist setting, but print a warn about this not allowed combination to inform about a possible misconfiguration.

src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
Andre601: Adressed this.

src/main/java/me/clip/placeholderapi/expansion/manager/LocalExpansionManager.java
Andre601: Moved it

BlitzOffline (approved):

PR #1114, opened 2025-04-30
[Wiki] Add Multiverse-Core v5 placeholders

Andre601:
### Type

- [ ] Other: __________ 

### Description

Closes N/A 

Multiverse-Core's v5 update implements direct PlaceholderAPI support through their own internal expansion.
As such is the expansion currently available on the eCloud becoming obsolete.

This Pull request adds the new placeholders, while keeping the old ones for documentation reasons.
It will be merged once MV releases a first stable v5 release (They are currently in pre-release stage).

I also suggest that the expansion should be unverified in the future and the MV expansion repo archived as their need will be no longer there.

### DIFF
--- a/docs/users/placeholder-list.md
+++ b/docs/users/placeholder-list.md
@@ -4203,9 +4203,46 @@ Miscellaneous placeholders:
 
 ----
 
-### **[Multiverse-Core](https://www.spigotmc.org/resources/390/)**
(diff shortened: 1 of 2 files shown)


Kqliber:
Will unverify the expansion and add a note on the ecloud site once v5 is released

PR #1172, opened 2026-02-08
Feat/add more hytale api docs

PiggyPiglet:
Add configurable api docs and also add common issue regarding threads & placeholder access in hytale

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -90,6 +90,7 @@ nav:
     - developers/index.md
     - developers/using-placeholderapi.md
     - developers/creating-a-placeholderexpansion.md
+    - developers/configuration-api-hytale.md
     - developers/expansion-cloud.md
   - Common Issues:
     - issues/index.md


docs/developers/configuration-api-hytale.md
```diff
@@ -0,0 +1,69 @@
...
+}
+```
+
+Your config class needs to represent your config structure exactly. The only difference is in naming, in the above example, dateFormat will be converted to snake case (date_format) when reading the yaml. The above class will look like this in PAPI's config:
+
+```yaml { .annotate title="HelpChat_PlaceholderAPI/config.yml" }
+expansions:
+	coolexpansion:
+		test: wew
+		date_format: "dd/mm/yyyy"
```
Andre601:
```suggestion
    coolexpansion:
        test: wew
        date_format: "dd/mm/yyyy"
```

Andre601 (changes requested):
One small correction for the Code example, to use spaces and not tabs for consistency.

Also, the page needs to be added to the `mkdocs.yml` file's `nav` section.

PR #1227, opened 2026-06-16
fix(parsing): adjust parsing to support paper dev builds better

### LINKED ISSUE #1226: 26.2 Support
### Confirmation

- [x] My issue isn't already found on the Issue tracker.
- [x] My issue is about **PlaceholderAPI** and not any expansion or external plugin.
- [x] The issue isn't already fixed in a Spigot Release or Development Build.
- [x] The [Common Issues](https://wiki.placeholderapi.com/common-issues/) page doesn't mention this issue.

### Type

Plugin Bug

### What happens?

[18:08:01 ERROR]: [ModernPluginLoadingStrategy] Could not load plugin 'PlaceholderAPI-2.12.2.jar' in folder 'plugins'
org.bukkit.plugin.InvalidPluginException: java.lang.ExceptionInInitializerError
        at io.papermc.paper.plugin.provider.type.spigot.SpigotPluginProvider.createInstance(SpigotPluginProvider.java:129) ~[paper-26.2.jar:26.2-10-5979742]
        at io.papermc.paper.plugin.provider.type.spigot.SpigotPluginProvider.createInstance(SpigotPluginProvider.java:35) ~[paper-26.2.jar:26.2-10-5979742]
        at io.papermc.paper.plugin.entrypoint.strategy.modern.ModernPluginLoadingStrategy.loadProviders(ModernPluginLoadingStrategy.java:116) ~[paper-26.2.jar:26.2-10-5979742]
        at io.papermc.paper.plugin.storage.SimpleProviderStorage.enter(SimpleProviderStorage.java:38) ~[paper-26.2.jar:26.2-10-5979742]
        at io.papermc.paper.plugin.entrypoint.LaunchEntryPointHandler.enter(LaunchEntryPointHandler.java:39) ~[paper-26.2.jar:26.2-10-5979742]
        at org.bukkit.craftbukkit.CraftServer.loadPlugins(CraftServer.java:549) ~[paper-26.2.jar:26.2-10-5979742]
        at net.minecraft.server.dedicated.DedicatedServer.initServer(DedicatedServer.java:297) ~[paper-26.2.jar:26.2-10-5979742]
        at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1290) ~[paper-26.2.jar:26.2-10-5979742]
        at net.minecraft.server.MinecraftServer.lambda$spin$0(MinecraftServer.java:302) ~[paper-26.2.jar:26.2-10-5979742]
        at java.base/java.lang.Thread.run(Thread.java:1474) ~[?:?]
Caused by: java.lang.ExceptionInInitializerError
        at java.base/java.lang.Class.forName0(Native Method) ~[?:?]
        at java.base/java.lang.Class.forName(Class.java:547) ~[?:?]
        at org.bukkit.plugin.java.PluginClassLoader.<init>(PluginClassLoader.java:82) ~[paper-api-26.2.build.10-alpha.jar:?]
        at io.papermc.paper.plugin.provider.type.spigot.SpigotPluginProvider.createInstance(SpigotPluginProvider.java:125) ~[paper-26.2.jar:26.2-10-5979742]
        ... 9 more
Caused by: java.lang.NumberFormatException: For input string: "b"
        at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67) ~[?:?]
        at java.base/java.lang.Integer.parseInt(Integer.java:565) ~[?:?]
        at java.base/java.lang.Integer.parseInt(Integer.java:662) ~[?:?]
        at PlaceholderAPI-2.12.2.jar//me.clip.placeholderapi.PlaceholderAPIPlugin.<clinit>(PlaceholderAPIPlugin.java:71) ~[?:?]
        ... 13 more

### Expected Behaviour

no error

### How to Reproduce

Starting Paper 26.2 Server with PlaceholderAPI 2.12.2

### `/papi dump` Output

/

### Console Log

_No response_

### Errors

_No response_

### Additional Info

_No response_

darbyjack:
### Type

- [ ] Other: __________ 

### Description

Change the way that version parsing is done to strip Paper dev build metadata in the version string.

Closes #1226  

### DIFF
--- a/src/main/java/me/clip/placeholderapi/ServerVersionResolver.java
+++ b/src/main/java/me/clip/placeholderapi/ServerVersionResolver.java
@@ -1,0 +1,133 @@
+/*
+ * This file is part of PlaceholderAPI
+ *
+ * PlaceholderAPI
+ * Copyright (c) 2015 - 2026 PlaceholderAPI Team
+ *
+ * PlaceholderAPI free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PlaceholderAPI is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package me.clip.placeholderapi;
+
+import java.util.OptionalInt;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+final class ServerVersionResolver {
+
+    private static final String PAPER_BUILD_INFO_PROVIDER =
+            "me.clip.placeholderapi.PaperServerBuildInfoProvider";
+
+    private ServerVersionResolver() {
+    }
+
+    @NotNull
+    static ServerBuild resolve(@NotNull final String bukkitVersion) {
+        return resolve(bukkitVersion, resolvePaperBuildInfo());
+    }
+
+    @NotNull
+    static ServerBuild resolve(@NotNull final String bukkitVersion,
+                               @Nullable final ServerBuild paperBuildInfo) {
+        if (paperBuildInfo != null) {
+            return paperBuildInfo;
+        }
+
+        return new ServerBuild(normalizeBukkitVersion(bukkitVersion), OptionalInt.empty());
+    }
+
+    @Nullable
+    private static ServerBuild resolvePaperBuildInfo() {
+        try {
+            final Class<?> provider = Class.forName(
+                    PAPER_BUILD_INFO_PROVIDER,
+                    true,
+                    ServerVersionResolver.class.getClassLoader()
+            );
+            final Object result = provider.getMethod("get").invoke(null);
+
+            if (result instanceof ServerBuild) {
+                return (ServerBuild) result;
+            }
+        } catch (final ReflectiveOperationException | LinkageError ignored) {
+            // Paper is unavailable or exposes an incompatible ServerBuildInfo API.
+        }
+
+        return null;
+    }
+
+    @NotNull
+    static String normalizeBukkitVersion(@NotNull final String bukkitVersion) {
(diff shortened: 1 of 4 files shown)


Andre601:
I feel like this could be simplified and improved on Paper Servers, by using their `ServerBuildInfo` class, if available, to retrieve and handle versions and build numbers.

Of course, this would require to keep it in a separate class to avoid exceptions for unrecognized imports on Spigot or older Paper forks, but I believe the benefits here would be useful.

More info on the ServerBuildInfo: https://jd.papermc.io/paper/26.2/io/papermc/paper/ServerBuildInfo.html

(That class existed for a few versions already btw.)

darbyjack:
> I feel like this could be simplified and improved on Paper Servers, by using their `ServerBuildInfo` class, if available, to retrieve and handle versions and build numbers.
> 
> Of course, this would require to keep it in a separate class to avoid exceptions for unrecognized imports on Spigot or older Paper forks, but I believe the benefits here would be useful.
> 
> More info on the ServerBuildInfo: https://jd.papermc.io/paper/26.2/io/papermc/paper/ServerBuildInfo.html

I like this idea and I think it's a better approach. I've pushed some more changes if you want to look it over and see if it's what you had in mind.

src/main/java/me/clip/placeholderapi/ServerVersionResolver.java
```diff
@@ -0,0 +1,133 @@
...
+                    PAPER_BUILD_INFO_PROVIDER,
+                    true,
+                    ServerVersionResolver.class.getClassLoader()
+            );
+            final Object result = provider.getMethod("get").invoke(null);
+
+            if (result instanceof ServerBuild) {
+                return (ServerBuild) result;
+            }
+        } catch (final ReflectiveOperationException | LinkageError ignored) {
```
Andre601, line 64: Are you sure this will catch Java throwing when trying to load a class containing imports to a non-existant object?

I myself had this once with Paper's Dialog API, hence why I suggested here to have it separate and do a check for the class in question.

OstlerDev:
I have tested this on Paper `26.2-23-dev` and can confirm that it does work and the plugin starts up successfully and works without issue. I cannot comment about the reviewers question about it properly catching errors for odd situations, but this does work as expected for me.

src/main/java/me/clip/placeholderapi/ServerVersionResolver.java
PiggyPiglet: Have tested on some historical versions and doesn't seem to cause any issues

PR #1237, opened 2026-07-01
Type Handlers

### LINKED ISSUE #1208: Allow placeholder to return text component
### Confirmation

- [x] I checked the Issues and Pull request tab for any existing issues or PRs.
- [x] My Feature Request is for **PlaceholderAPI** and not any expansion or other plugin.

### Type

New API feature

### Description

This is a suggestion to allow placeholders to return adventure components. Text components are more and more powerful with the new recent features.

With components, you can make placeholders that return text with hover, text with custom font, return sprites (glyph) player heads and more.

### Example use cases:
-  Leaderboard placeholders that return the player head of the player
-  Nexo placeholders currently return `<glyph:id>`, then verify every packet and replace the value by a custom font & chars, they are using a custom font to not replace Minecraft characters to make glyph. Allowing to return components can be way more optimized and clean.
-  A recent plugin was released that allows to create rank tags without a texture pack, using only player heads and custom skins (mineskin) to create images without a texture pack. Without component support, it’s not possible for that plugin to add PAPI support.

### Example implementation:
-  This is already in MiniPlaceholders, having it in PlaceholderAPI will allow for way more plugin compatibility. I know that plugins will need to update their code to support it, but it’s still a community that’s way bigger. Server owners also don't want to have two plugins to do almost the same thing.

### Code Example

```java

```

### Jar file

_No response_

### Additional Information

_No response_

PiggyPiglet:
### Type

- [ ] Other: __________ 

### Description

Closes #1208

This is an idea from [PAPI3](https://github.com/PlaceholderAPI/PlaceholderAPI3/issues/30), however the implementation in this version doesn't exactly match what it would've in papi 3.

A developer could for example have something like
```java
case "loc":
        return player.getLocation();
```

and should a type handler for Location be registered, that type handler could then provide placeholders like _x, _y, etc, allowing for a standarised format and predictable results across all expansions returning a specific type versus independent implementations that may differ.

I'm not happy with the implementation or api design here... at all, but I only had 2 hours to work on this so that's where we're at. In a perfect world a type handler would attach directly to an expansion and for example %player_x% would work, wouldn't need %player_loc_x%, however that's out of scope for tonight.

### DIFF
--- a/src/main/java/me/clip/placeholderapi/PlaceholderAPI.java
+++ b/src/main/java/me/clip/placeholderapi/PlaceholderAPI.java
@@ -65,8 +65,8 @@ public final class PlaceholderAPI {
     @NotNull
     public static String setPlaceholders(final OfflinePlayer player,
                                          @NotNull final String text) {
-        return REPLACER_PERCENT.apply(text, player,
-                PlaceholderAPIPlugin.getInstance().getLocalExpansionManager()::getExpansion);
+        final LocalExpansionManager manager = PlaceholderAPIPlugin.getInstance().getLocalExpansionManager();
+        return REPLACER_PERCENT.apply(text, player, manager::getExpansion, manager::resolvePlaceholderValue);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


Andre601:
Why the rush? Just wait with making this in all honesty.

PiggyPiglet:
Not really any rush, doubt this will be ready for next release

PR #1249, opened 2026-09-12
Handle expansion errors during placeholder parsing

### LINKED ISSUE #1107: Placeholder parsing does not fail gracefully when exception is thrown by expansion
### Confirmation

- [x] My issue isn't already found on the Issue tracker.
- [x] My issue is about **PlaceholderAPI** and not any expansion or external plugin
- [x] The issue isn't already fixed in a Spigot Release or Development Build.
- [x] The [Common Issues](https://github.com/PlaceholderAPI/PlaceholderAPI/wiki/Common-Issues) page doesn't mention this issue.

### Type

Plugin Bug

### What happens?

If an expansion throws or fails to catch an error during the parsing process, the error is not caught by PlaceholderAPI and is propagated further.

### Expected Behaviour

PlaceholderAPI should not assume that the expansion code is exception-proof. It should catch any exception, log the error and continue its work.

Surround the following code line with a try-catch block and in the case an exception is caught, execute the same behavior as if no expansion was found with an extra log message: https://github.com/PlaceholderAPI/PlaceholderAPI/blob/e019d6370c5b0f61f5c86a4b5bfe276aa6f6ebc4/src/main/java/me/clip/placeholderapi/replacer/CharsReplacer.java#L119

### How to Reproduce

Create a simple expansion that throws an exception in the onRequest/onPlaceholderRequests methods and attempt to parse a placeholder using the PlaceholderAPI.setPlaceholder method.

### `/papi dump` Output

NO!

### Console Log

_No response_

### Errors

_No response_

### Additional Info

Here I have created a simple expansion that intentionally throws a RuntimeException no matter what the parameters are: https://paste.helpch.at/udexuzoraj.typescript

Image

Image

HP-network:
A throwing expansion currently aborts the entire placeholder parse. Catch expansion exceptions, log the failing placeholder through the expansion logger, and preserve the original token so later placeholders can still be processed.

Adds a regression test for the failure path. Fixes #1107.


Andre601:
Personally, I feel like it's the responsibility of the expansion to properly handle and deal with exceptions.

From the PoV of PlaceholderAPI is it not expected for an expansion to throw anything on parse, but rather that this is handled gracefully and either a String or null being returned.
Having an expansion throw on parse, breaking the parsing tells me, that this is a failure on the expansion and that it should be reported to whoever made the expansion to get it fixed.

This genuinly feels like a band-aid solution to me.
Maybe PiggyPiglet or Funnycube think different tho.

HP-network:
Thanks for the feedback. I agree that an expansion should handle its own exceptions, but I think PlaceholderAPI still needs to treat expansion callbacks as a third-party boundary.

`onRequest` is called from `CharsReplacer`, and an unchecked exception currently aborts the entire parse. That means one broken expansion can prevent unrelated placeholders in the same string from being resolved; this is the failure described in #1107.

This change does not hide the failure: it logs the exception through the expansion logger, leaves the original placeholder unchanged (the same fallback used when no replacement is returned), and catches `Exception` rather than `Error`. The intent is containment, not replacing the need to fix the expansion itself.

Would you prefer a different logging or fallback behavior, or handling this at another layer?

